Issue #14, opened 2018-11-13
Language support in FHIR API


goranoe:
As far as I can see the FHIR API does not support any other languages than the language imported in the MAIN branch. I am running Snowstorm with the international release in the MAIN branch and the Swedish edition in MAIN/SNOMEDCT-SE. I've tested the FHIR lookup-operations on CodeSystem and it always uses the MAIN branch, so I am not able to get the Swedish translations through FHIR . The regular REST-API works using the Accept-Language http-header, so I have verified that the Swedish release has been imported correctly.

rorydavidson:
Both the FHIR coverage and implementation are experimental on snowstorm at the moment and I'd not recommend using it yet. This is on our roadmap in the first half of next year, although we always welcome any contributions to improve this in the meantime.

goranoe:
Thanks for the quick response! I will look at it and try to contribute.

kaicode:
Included in release 2.2.0.

PR #15, opened 2018-11-20
Issue #14 Language support in FHIR API

goranoe:
I created an issue (https://github.com/IHTSDO/snowstorm/issues/14) about FHIR API not having support for different language editions. The API always used the MAIN branch. 

Local editions are now supported when imported under main branch when branch name given as Snomed SCTID in accordance with https://confluence.ihtsdotools.org/display/DOC/List+of+SNOMED+CT+Edition+URIs
I.e. MAIN/45991000052106 for Swedish edition

The MAIN branch is expected to contain the International edition. Choosing which Snomed edition to query is done using the FHIR parameter "version" in accordance with the FHIR standard.


kaicode:
Thank you for the contribution @goranoe! Peter will aim to review this in the next few days.

goranoe:
I've found there's an improvement still needed in HapiCodeSystemMapper.java in method addDesignations where there is still a hard-coding for US English. I am working on a fix for that.

pgwilliams:
Hi @goranoe, that was really great code - nicely written and very much appreciated!    

I was forced to make a tweak because - internally - we name extension branches use a code system shortname like SNOMEDCT-SE rather than the module id.    In fact using the module id makes more sense because it's not arbitrary but when you're looking through many branches or typing in URLs the short name is more human friendly.   We'll have a talk about that at our end and see if we can avoid that hardcoding - or perhaps some server level configuration for a branch naming strategy.

Thanks again & Best Wishes, Peter

kaicode:
I have refactored this implementation
- Moved the code system module information into the configuration file
https://github.com/IHTSDO/snowstorm/blob/dea4b2648ac936588cd42362fa0f26249fdb789d/src/main/resources/application.properties#L102
- Allowed code systems and version branches to by found using the Snowstorm CodeSystem service.
- Used the Accept-Language request header to select the best language to return rather than using extension defaults. This is inline with the rest of Snowstorm.

PR #23, opened 2019-02-05
Add the maven wrapper

kevinbayes:
Update to the maven build to include the wrapper. This ensures that the exact version of maven is available for building.

Update to the .gitignore to allow the following files and folders to be added:
* .gitignore
* .travis.yml
* .mvn/


kaicode:
Hi @kevinbayes I find this maven wrapper quite disruptive. If we wanted to enforce a minimum maven version programmatically the [maven-enforcer-plugin](https://stackoverflow.com/a/32971779/512223) looks like a good option.
Cheers.

kevinbayes:
Fair enough. I don't like installing maven on my path so thought it would be helpful.

PR #24, opened 2019-02-05
Update dockerfile

kevinbayes:
Update the Dockerfile to run under a user. 

Add a docker-entrypoint.sh to ensure that future proof for any complex startup that may be required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,29 @@
 FROM openjdk:8-jdk-alpine
 
+ARG SUID=1042
+ARG SGID=1042
+ENV PROJ_VERSION '2.1.0'
+ENV PROJ_JAR "snowstorm-$PROJ_VERSION.jar"
+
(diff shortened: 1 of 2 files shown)


kevinbayes:
Please let me know what I should do further to enhance this to get merged. Like perhaps removing the need for the version in the dockerfile? Unless there is a specific reason it is there.

kaicode:
Hi @kevinbayes. Thanks for starting this. Could we remove the .sh script until it's needed please? Let's strive for simplicity until the requirements prevent it. The java command in the docker file could be updated inline if/when needed.
If you were able to remove the need for the version number in the docker file that would be great. I expect we can use a wildcard in the ADD command. That would be preferable to changing the name of the jar that the maven build produces. 
Thank you.

kevinbayes:
Done, let me know if there is anything else.

PR #30, opened 2019-03-04
Add configurable max window size

iCart:
Hello,
This makes the maximum page size configurable.
This is useful when you need to get more than 10000 results from an ECL query for example. (provided you also increase the max window size in the relevant indexes in Elasticsearch).

Another workaround would be to instead use Elasticsearch's scroll API in Snowstorm, but that seems like overkill since that doesn't happen that often.


kaicode:
@iCart Thanks for the idea. We have actually implemented searchAfter functionality in the develop branch. I'm hesitant to accept this because people will not be guaranteed to get all their search results back unless they increase the max window size to many times the recommended value.

Would using search after work for you? Each page has a searchAfter token which you can use to fetch the next page. This has no limit to the number of results you can fetch and has no impact on Elasticsearch memory or performance.

iCart:
@kaicode Oh, i didn't realise the searchAfter functionality was already in the develop branch.
After a bit of testing, it does look like it fits our needs.

I'll close this PR then (for the reasons you raised and because it's not needed anymore), thanks!

PR #38, opened 2019-04-16
STORM-184: Add transientEffectiveTime option during RF2 export

CoderMChu:
Hi Kai,

Can you review my changes please?

Thanks

Michael


CoderMChu:
That is taken care of in the ExportWriter.(i.e The default value of transient effective time is blank).

PR #46, opened 2019-05-07
Fix loading-snomed doc link

me2resh:
Current loading snomed link in using-docker documentation is broken

Fixed the reference for it to go to
https://github.com/IHTSDO/snowstorm/blob/master/docs/loading-snomed.md
instead of 
https://github.com/IHTSDO/snowstorm/blob/master/docs/docs/loading-snomed.md


kaicode:
Thanks @me2resh ! 😺

PR #52, opened 2019-06-05
Batch extract mode

gemanilkashyap:
Requirement - Extract concepts with no limitation with satisfying criteria passed as ECL. Offline processing done to extract more information needed for business purpose.

Eg : Need to extract all UK clinical products

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/QueryService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/QueryService.java
@@ -14,6 +14,7 @@ import org.snomed.snowstorm.core.data.services.pojo.ResultMapPage;
 import org.snomed.snowstorm.core.util.PageHelper;
 import org.snomed.snowstorm.core.util.TimerUtil;
 import org.snomed.snowstorm.ecl.ECLQueryService;
+import org.snomed.snowstorm.rest.ControllerHelper;
 import org.snomed.snowstorm.rest.converter.SearchAfterHelper;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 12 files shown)


kaicode:
Hi @gemanilkashyap, thanks for suggesting a contribution.
Could I clarify your requirements please? It looks like you would like the ability to download an unlimited number of concepts from an ECL query? Is that all?

kaicode:
We have a new feature for the concept search using the _searchAfter_ parameter. This allows paging through an unlimited amount of results. Using _offset_ has a limit of 10K results but _searchAfter_ has no limit.

1. Start your concept search as normal using the _{branch}/concepts_ endpoint.
2. The response will contain a searchAfter value.
3. Request the next page by including the searchAfter request parameter using the searchAfter value from the previous response.
4. Repeat steps 2 and 3 until the end of the results where you will get a blank page back.

Apologies for lack of documentation. I will endevor to add this to the docs in the next release.
Give it a try and let us know if that satisfies your requirements!

gemanilkashyap:
Sure @kaicode thanks for suggesting, yeah the requirement is to pass beyond the limitation to get all concepts satisfying the criteria in query. I will check with the above option and revert back.

gemanilkashyap:
@kaicode i am not getting searchAfter attribute as part of response when i ran
http://localhost:8080/MAIN/concepts?term=asthma
Am i missing something?

kaicode:
You should get searchAfter at the end of the response. 
The response should look like:
```
{
"items": [......],
"total": 967,
"limit": 50,
"offset": 0,
"searchAfter": "WzE0LDE0LjU5NzkwNV0=",
"searchAfterArray": [
14,
14.597905
]
}
```
Please check you have pulled and built the latest version of Snowstorm.
http://localhost:8080/version should show version `3.0.3` (later if you are on the develop branch).

gemanilkashyap:
I was expecting the same but my response was 

total | 1359
-- | --
limit | 50
offset | 0
Only
Was this feature introduced in latest code i mean from past few days, then i might have to redeploy..

kaicode:
This feature was introduced in version `3.0.3` which was pushed to master 19 days ago on May 17th
https://github.com/IHTSDO/snowstorm/releases/tag/3.0.3
http://localhost:8080/version will tell you which version you have.

gemanilkashyap:
Apologies it looks like i am using old version, will upgrade and test now to 3.0.3

gemanilkashyap:
@kaicode I reloaded the data with latest codebase and it looks like working from browser. I need to change my batch job to reflect the changes so will revert back once loaded the whole set of concepts. Also does the limit of 50 will affect performance when doing a batch extract?

kaicode:
With batch loading of content in general making less round trips should be faster so using large pages of around 500 concepts at a time should work well. Play around with different page sizes to see what works best in your environment.

kaicode:
Closing this one because we have an alternative solution already implemented. Thank you to @gemanilkashyap for being proactive with this all the same!

PR #82, opened 2019-11-08
Make browser/{branch}/concepts/bulk-load visible in read-only mode

metaruslan:
Currently all POST/PUT/PATCH/DELETE are inaccessible in read-only mode, but the "/browser/{branch}/concepts/bulk-load" endpoint is a POST, so it cannot be accessed even though it's not changing any state. My fix made it a GET instead of a POST. This fix can potentially break existing clients, because the conceptIds are a GET parameter now, not a POST body parameter, but it is actually more consistent with the existing "/{branch}/concepts" endpoint that also takes the conceptIds as a GET parameter.


kaicode:
Hi @metaruslan, thanks for raising this problem and getting the ball rolling!

The reason we have that endpoint as a POST is due to the request size restriction on GET requests. A GET request is limited to 2,048 characters because it's all got to fit within an HTTP URL whereas the size of a POST request does not have this limit. Using POST many thousand concepts can be requested at once.

I notice that the equivalent GET request for browser concepts does not accept the conceptIds parameter. That is an oversight on my part, let me add that as an alternative approach.

metaruslan:
@kaicode, thanks for pointing out on the GET parameter size limitation, I actually missed it! Sounds like another approach is needed. You are going to add conceptIds to the other GET method, but that will still have this limitation.
Here is my other suggestion - leave the bulk-load endpoint as is and modify
@Bean
	public List<String> allowReadOnlyPostEndpointPrefixes() {
		return Collections.singletonList("/fhir");
	}
to make an exception for the endpoint. What do you think?

metaruslan:
@kaicode, actually I don't see a good prefix that would isolate just the "/browser/{branch}/concepts/bulk-load" endpoint, so probably we need to hardcode just this endpoint? I can create another pull request if it sounds correct.

kaicode:
The other issue with enabling the POST for read-only mode would be that sometimes people try to load most of SNOMED from our public browser.. I need to tread a fine line here, would the GET version of this endpoint be enough for your use case? Otherwise we may need to add another flag which people can set in their own deployments to allow the POST endpoint you are suggesting..

metaruslan:
No, GET won't be enough for our team. We are specifically trying to optimize querying our own instance of snowstorm and having the ability to load a big number of concepts in one pass is what we are looking for. Right now we are left with disabling read-only mode. Having the suggested flag would work well for us.

metaruslan:
@kaicode thank you for the fix for the GET method. As for the opening up the POST bulk-load with a flag - do you want me to create another pull request or do you want to implement it your way?

kyrill007:
Well, Ruslan, we can page in 100 ids increments if this gets us the fix faster. :)

kaicode:
Sorry @metaruslan, I'm travelling and have not responded for a few days. Yes, please do create a pull request with a flag to allow POST bulk-load 👍 That would be very helpful.

metaruslan:
@kaicode thank you! The code is ready, but I'd like my colleagues to do a review before I create the pull request. Probably I'll create it on Monday.

metaruslan:
@kaicode I've created a new pull request:
https://github.com/IHTSDO/snowstorm/pull/85
Closing this one. Thanks a lot!

PR #85, opened 2019-11-15
Make browser/{branch}/concepts/bulk-load accessible in read-only mode…

metaruslan:
… when a newly introduced flag is enabled.

This is the result of the decision made in this discussion:
https://github.com/IHTSDO/snowstorm/pull/82

We need to make the "browser/{branch}/concepts/bulk-load " endpoint accessible in read-only mode, but we need to introduce a new flag for it, because the Snowstorm team doesn't want this endpoint to be accessible in the public server even though it's read-only, because sometimes people try to load most of SNOMED from their public browser. Thus this flag is needed to make the switch for private installations.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/config/SecurityAndUriConfig.java
+++ b/src/main/java/org/snomed/snowstorm/config/SecurityAndUriConfig.java
@@ -51,6 +51,9 @@ public class SecurityAndUriConfig extends WebSecurityConfigurerAdapter {
 	@Value("${snowstorm.rest-api.readonly}")
 	private boolean restApiReadOnly;
 
+	@Value("${snowstorm.rest-api.readonly.allowReadOnlyPostEndpoints}")
+	private boolean restApiAllowReadOnlyPostEndpoints;
+
 	@Value("${ims-security.roles.enabled}")
 	private boolean rolesEnabled;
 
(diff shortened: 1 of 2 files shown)


metaruslan:
@kaicode thank you so much!

metaruslan:
@kaicode my team is curious about the timeline you guys are planning for a release that will include this feature.

kaicode:
A release will be cut this week for some Snomed browser enhancements.

kaicode:
@metaruslan This pull request has been tested and included in 4.5.0.
https://github.com/IHTSDO/snowstorm/releases/tag/4.5.0

Thank you for your contribution 🙇

metaruslan:
Thanks a lot! It has been a pleasure contributing to this project :-)

metaruslan:
Hi @kaicode 

Funny we just found a bug in it :-) It doesn't let the branch names that contain "/" to pass thru, i.e. the ant matcher considers it to be a different path. I did test my code, but without such a branch name, I assume you did the same in your testing. So, I'm going to come up with a fix when I find some time and make the matcher accept:
anything that starts with "/browser/" and ends with "/concepts/bulk-load"
Let me know if such a pull request may not be accepted. Or maybe you guys could have made it yourself by the time I get to it, that would also work.

Thanks in advance!

metaruslan:
Probably not the best place to report a bug as this is a closed pull request. I have just created an issue
https://github.com/IHTSDO/snowstorm/issues/95
to keep track of it.

PR #89, opened 2019-11-25
Update loading-snomed.md

JonZammit:
Changed to "Xmx2g" to be consistent with https://github.com/IHTSDO/snowstorm/blob/master/docs/getting-started.md


kaicode:
Well spotted Jon. I'll update the getting started one to max 4g instead.

kaicode:
Have updated in the develop branch:
https://github.com/IHTSDO/snowstorm/blob/develop/docs/getting-started.md

PR #96, opened 2019-12-20
Adding FHIR Medication Resource endpoint

alopezo:
This a new endpoint in the FHIR API (/fhir/Medication) that retrieves a Medication resource created form the definition of a given SNOMED CT Clinical Drug Concept.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/fhir/config/HapiRestfulServlet.java
+++ b/src/main/java/org/snomed/snowstorm/fhir/config/HapiRestfulServlet.java
@@ -7,6 +7,7 @@ import javax.servlet.ServletException;
 
 import org.snomed.snowstorm.fhir.services.FHIRCodeSystemProvider;
 import org.snomed.snowstorm.fhir.services.FHIRConceptMapProvider;
+import org.snomed.snowstorm.fhir.services.FHIRMedicationProvider;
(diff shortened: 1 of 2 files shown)


pgwilliams:
Awesome contribution, thank you so much Alejandro!

alopezo:
Yes @kaicode , you are right. The FHIR API provides the challenge of how to send additional snowstorm or SNOMED specific parameters like path and language for example. Not sure what was your approach for other FHIR endpoints, probably using headers is less intrusive over the standard FHIR API?

kaicode:
Hi @alopezo,

I'm still getting to grips with the FHIR specification but on the Snowstorm side if no path is given we should default to the branch of the latest release for the codesystem in use. So for International right now that would be MAIN/2020-01-31. This is because we typically host daily build content on MAIN. 
In order to select a different code system or branch I guess you would use the URI of a different edition in the FHIR request. Again this is a bit fuzzy but we started building a [map of code systems and modules](https://github.com/IHTSDO/snowstorm/blob/4.7.1/src/main/resources/application.properties#L152) for the purposes of this lookup but I don't think it's being used anywhere yet. 

Peter and I will be pairing on the development of the FHIR API in Snowstorm in the next couple of weeks so hopefully we can implement some of these ideas in the endpoints we have to provide some clear direction.
Thanks again for your contribution, we really appreciate it.

Cheers, Kai

PR #108, opened 2020-04-08
MAINT-1027 Fetch first 10K results for branch children API endpoint and Paging.

kparmar1:


kaicode:
We have a large number of commits in this request. I changed the base to the develop branch but that didn't help. Could you rebase your develop branch against the IHTSDO develop branch to see if that help please @kparmar1 ?

danka74:
FWIW, I have built some code around the fact that snowstorm paginates results using rxjs observables. It works fine but I see that performance is increased significantly by lowering page size to e.g. 1k compared to 10k. When doing large-ish tasks as translating every defined medicinal product (20k concepts) it is the difference between a 30 min job and not finishing over night.

kparmar1:
Sorry this PR was incorrect. It's against the wrong branch.

kparmar1:
@danka74 Thanks for the info. I will add this to the new PR.

PR #109, opened 2020-04-08
MAINT-1027 Fetch first 10K results for branch children API endpoint and Paging.

kparmar1:


kparmar1:
Comment by @danka74 

FWIW, I have built some code around the fact that snowstorm paginates results using rxjs observables. It works fine but I see that performance is increased significantly by lowering page size to e.g. 1k compared to 10k. When doing large-ish tasks as translating every defined medicinal product (20k concepts) it is the difference between a 30 min job and not finishing over night.

kparmar1:
@kaicode Changes made to the names. The regex should be good to match as it already accounts for everything after the '/' but not a '/'.

The tests after the change were fine.

kaicode:
Ah, I see now. The regex could be simplified to `[^/]*`, I think that's what confused me. Looks good. Will merge.

kparmar1:
@kaicode Thanks.

PR #123, opened 2020-05-28
Updated the baseUrl

aHoejen:
I have updated this file to include /fhir in the baseUrl. This is more consistent with the specification and other implementations.


kaicode:
This looks good to me. Could you take a peek too please @pgwilliams ?

pgwilliams:
Yup, looks great, thanks @aHoejen !

PR #127, opened 2020-06-12
Develop

aHoejen:
Not sure I am doing this right, as I couldn't push to origin in Github Desktop. - please advise if this is not correct. However, the changes I have made to the file should be the following:

- changed the baseUrl to the training server, as this will be what the students will be using.
- Added a single line version of all requests to support copy-pasting
- Adding a number to each of the requests (in the comments), to support communication and referencing
- removed one instance of system-version (R7), as this can be applied to the url parameter


rorydavidson:
This pull request is being retracted

PR #128, opened 2020-06-12
Develop

aHoejen:
- Changed the baseUrl to the training server, as this will be what the students will be using.
- Added a single line version of all requests to support copy-pasting
- Adding a number to each of the requests (in the comments), to support communication and referencing
- Removed one instance of system-version (R7), as this can be applied to the url parameter


rorydavidson:
Retracting

PR #134, opened 2020-06-25
Update fhir-requests.http

aHoejen:
Update to the newest version of the file. 
Removing version parameter from R15.


rorydavidson:
Thanks @aHoejen

PR #146, opened 2020-08-13
Elasticsearch 7.7.0 upgrade with spring data 

CoderMChu:
Hi Kai,

The Elasticsearch 7.7.0 upgrade is ready for code review and dev testing.

Thanks
Michael


kaicode:
SonarCloud thinks this is a bug, it may ignore it if the return type in the method signature matches the declared type of the local variable or if the return statement is simplified.
https://sonarcloud.io/project/issues?id=IHTSDO_snowstorm&open=AXPoXizHt5xwImcWBVK7&pullRequest=146&resolved=false&types=BUG

CoderMChu:
> This looks great @CoderMChu.
> 
> I've scanned through the mapping and service code changes and no concerns. The fact that you have this working without any functional unit test changes gives me a lot of confidence. The concept serialisation test has not needed to change which is good too.
> 
> I noticed that the Mixin classes were removed from the Elasticsearch config. Are they no longer needed? You have added some more Transient annotations. Some fields are needed in requests and responses in the REST interface but should not be serialised to Elasticsearch documents. It would be worth checking that we have not lost and fields in the REST interface with these changes. The concept serialisation test should cover this but I would like to understand how this is now working.

These Mixin classes are only used in the rest response now. They have been removed in the model config due to spring data elasticsearch 4 is no longer supporting Jackson based mapper. Instead it uses meta model object mapping https://docs.spring.io/spring-data/elasticsearch/docs/current/reference/html/#elasticsearch.mapping.meta-model

Yes we need to check fields are not lost in the rest response.

PR #187, opened 2020-11-24
CDI-29 - Add support for importing concrete Relationships

dmcgihtsdo:
CDI-29 is concerned with adding support for importing concrete Relationships.

To test this:
- Import a Daily Build containing concrete data
- Navigate to both a normal Relationship and a concrete Relationship on Snowstorm (i.e. "http://localhost:8080/$branch/relationships/$relationshipId")
- For a normal Relationship, the 'destinationId' field will be populated whilst the 'value' field will not be returned
- For a concrete Relationship, the 'value' field will be populated whilst the 'destinationId' field will not be returned


src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
```diff
@@ -129,7 +136,17 @@ public Relationship(String id, Integer effectiveTime, boolean active, String mod
 		this.active = active;
 		setModuleId(moduleId);
 		this.sourceId = sourceId;
-		this.destinationId = destinationId;
+
+		if (destinationId != null) {
+			if (destinationId.startsWith("#") || destinationId.startsWith("\"")) {
```
kaicode, line 141: I'm not sure about this but we may be able to rid of it later.

kaicode: We should at least name the parameter as destinationIdOrValue until we know what it is.

kaicode (approved):
Looks good to me. Thanks Darryn! Feel free to merge this once dev is working.

PR #190, opened 2020-11-30
CDI-28 - Return Object instead of String for concrete values

dmcgihtsdo:
CDI-28 is concerned with updating the response of Snowstorm to display concrete values as Objects rather than Strings. 

When requesting data of a normal Relationship, the fields "destinationId" and "target" will be populated. However, when requesting data of a concrete Relationship, the field "concreteValue" will replace both "destinationId" and "target". 

To test this:

- Ensure you have imported concrete data
- Navigate to '/$branch/relationships/$relationshipId' for both normal and concrete Relationships. You will either see "destinationId" and "target" fields or "concreteValue" fields.
- Navigate to '/$branch/concepts/$conceptId' for both normal and concrete Relationships. You will either see "destinationId" and "target" fields or "concreteValue" fields.


src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
```diff
@@ -0,0 +1,66 @@
...
+import org.snomed.snowstorm.rest.View;
+import org.springframework.data.annotation.Transient;
+
+/**
+ * Represent the Value of a Concrete Relationship
+ * in API response.
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+@JsonView(value = View.Component.class)
+public final class ConcreteValue {
```
kaicode, line 14: No need to mark classes as final. In general Snowstorm is not used as a library but we know that some implementers do extend it for wrapping into an OSGi bundle and other purposes, we don't want to block them extending our classes for whatever reason.

src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
```diff
@@ -0,0 +1,66 @@
...
+
+        if (!isConcreteNumber && !isConcreteString) {
+            throw new IllegalArgumentException(value + " is not a concrete value.");
+        }
+
+        if (isConcreteString) {
+            this.dataType = DataType.STRING;
+        } else {
+            value = value.substring(1);
+            this.dataType = value.contains(".") ? DataType.DECIMAL : DataType.INTEGER;
```
kaicode, line 53: This is a good fallback for determining the data type but we should use the MRCM primarily. I'll create a separate ticket for that.

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
```diff
@@ -62,11 +62,16 @@ public String getConceptId() {
 	@Size(min = 5, max = 18)
 	private String destinationId;
 
-	@JsonInclude(JsonInclude.Include.NON_EMPTY)
-	@JsonView(value = View.Component.class)
 	@Field(type = FieldType.Keyword)
+	@Size(min = 2, max = 4096)
```
kaicode, line 66: Great thanks. Min of 2 including the hash or open quotes makes sense 😄

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
```diff
@@ -62,11 +62,16 @@ public String getConceptId() {
...
+	@JsonIgnore
```
kaicode, line 67: You shouldn't need this. The absence of `@JsonInclude(JsonInclude.Include.NON_EMPTY)` should prevent it from being included in API responses and we want it included when writing to Elasticsearch.

src/main/java/org/snomed/snowstorm/core/data/services/AxiomConversionService.java
```diff
@@ -102,12 +102,29 @@ private AxiomRelationshipConversionService setupConversionService(String branchP
...
+					final Relationship internalRelationship = new Relationship(externalRelationshipTypeId + "", externalRelationshipDestinationId + "");
+					internalRelationship.setSourceId(sourceId != null ? sourceId.toString() : null);
+					internalRelationship.setGroupId(externalRelationship.getGroup());
+					internalRelationship.setInferred(false);
+
+					final org.snomed.otf.owltoolkit.domain.Relationship.ConcreteValue externalConcreteValue = externalRelationship.getValue();
+					if (externalConcreteValue != null) {
+						final String rf2Value = externalConcreteValue.getRF2Value();
+						internalRelationship.setConcreteValue(rf2Value);
+						internalRelationship.setValue(rf2Value);
```
kaicode, line 121: The code would be more robust if we could set both representations of the concrete value and clear destination and target with a single method call to the relationship object. See what you can do.

src/test/java/org/snomed/snowstorm/core/data/domain/ConcreteValueTest.java
```diff
@@ -0,0 +1,136 @@
...
+        final ConcreteValue.DataType result = concreteValue.getDataType();
+
+        //then
+        assertEquals(ConcreteValue.DataType.INTEGER, result);
+    }
+
+    @Test
+    public void concreteValue_ShouldStoreCorrectValue_WhenGivenConcreteString() {
+        //given
+        final String input = "\"\"A Promised Land\" is a memoir by Barack Obama.\"";
```
kaicode, line 84: Nice test case with quoted quotes 👍

src/test/java/org/snomed/snowstorm/core/data/domain/ConcreteValueTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+        //when
+        final ConcreteValue concreteValue = new ConcreteValue(input);
+        final String actualResult = concreteValue.getValue();
+        final String expectedResult = "3.14";
+
+        //then
+        assertEquals(expectedResult, actualResult);
+    }
+}
```
kaicode, line 136: Great unit test.

Please add a newline at the end of all classes to prevent these github warnings and make some types of file merges easier.

src/test/java/org/snomed/snowstorm/rest/ConceptSerialisationTest.java
```diff
@@ -126,33 +126,63 @@ void testStoreSerialisation() throws JsonProcessingException {
...
-        final String result = storeObjectMapper.writeValueAsString(relationship);
-        final boolean isConcrete = result.contains("value");
-        final boolean isOld = result.contains("destinationId");
-
-        //then
-        assertTrue(isConcrete);
-        assertFalse(isOld);
-    }
+	@Test
+	public void writeValueAsString_ShouldReturnCorrectString_WhenWritingOldDomain() throws JsonProcessingException {
```
kaicode, line 130: Thanks for correcting indentation.

kaicode (commented):
Looks good. A couple of small changes suggested.

src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
dmcgihtsdo: Fixed.

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
dmcgihtsdo: Fixed.

src/test/java/org/snomed/snowstorm/core/data/domain/ConcreteValueTest.java
dmcgihtsdo: Fixed.

src/main/java/org/snomed/snowstorm/core/data/services/AxiomConversionService.java
dmcgihtsdo: Fixed. 

I originally debated this myself, so I'm happy with the suggestion.

dmcgihtsdo:
I have made those changes @kaicode; open for a second look.

kaicode (approved):
Looks good to me! Please merge to develop when convenient.

PR #191, opened 2020-12-03
MAINT-1233 - Added support for traceability in RF2 delta imports.

Tyler-Cooksey:
Additions:

Refactored code to make it more readable in places.


src/main/java/org/snomed/snowstorm/core/data/services/TraceabilityLogService.java
```diff
@@ -65,18 +75,83 @@ public TraceabilityLogService() {
...
+	 * or deleted, both of these fields are classified as {@code Transient} so
+	 * the states will not be persisted when retrieved from the data store.
+	 * In the case where this method is called implicitly through the
+	 * {@link CommitListener}, by performing a {@link RF2Type#DELTA} import,
+	 * it will load the components from the data store so the {@code changed}
+	 * and {@code deleted} states will not be persisted which means that the
+	 * <code>useChangeFlag == false</code> so that other checks are ignored when
+	 * doing the comparison between what has changed. It's worth noting when doing
+	 * a {@link RF2Type#DELTA} import, it will show all the changes inside the
+	 * traceability log.
```
kaicode, line 112: It would be better to keep comments as brief and to the point as possible. Very long comments may be ignored and are likely to become out of date, especially when some of the information in them is about the workings of another class. The key information here is: "This method may be called by a service class as an optimisation. In this case the transient change and delete flags may be populated."

src/main/java/org/snomed/snowstorm/core/data/services/TraceabilityLogService.java
```diff
@@ -65,18 +75,83 @@ public TraceabilityLogService() {
...
```
kaicode, line 135: Same as above.

src/main/java/org/snomed/snowstorm/core/data/services/TraceabilityLogService.java
```diff
@@ -65,18 +75,83 @@ public TraceabilityLogService() {
...
 
 	@Override
-	public void preCommitCompletion(Commit commit) throws IllegalStateException {
-		if (commit.getCommitType() != CONTENT) {// Only use for rebase / promotion
-			logActivity(SecurityUtil.getUsername(), commit, new PersistedComponents());
+	public void preCommitCompletion(final Commit commit) throws IllegalStateException {
+		// Most commits don't use this because ConceptService calls logActivity directly.
+
+		PersistedComponents persistedComponents = null;
+		String importType = null;
```
kaicode, line 82: Perhaps this should be named `commitPrefix`?

src/main/java/org/snomed/snowstorm/core/rf2/rf2import/ImportService.java
```diff
@@ -25,7 +36,9 @@
 @Service
 public class ImportService {
```
kaicode, line 37: Nice refactoring of this class although be wary of repeating lengthy java doc descriptions when it's a private method anyway. A single comment somewhere may be cleaner?

src/test/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceTest.java
```diff
@@ -127,13 +141,59 @@ public Activity getActivityWithTimeout(int maxWait) throws InterruptedException
...
+	void testDeltaImportWithOneAdditionChange() {
+		activitiesLogged.clear();
+		final Commit commit = new Commit(branchService.create("MAIN/RF2DeltaImport"), Commit.CommitType.CONTENT, null, null);
+		final PersistedComponents persistedComponents =
+				PersistedComponents.newBuilder()
+								   .withPersistedConcepts(Collections.singleton(new Concept("3311481044").addFSN("Test FSN")))
+								   .withPersistedDescriptions(Collections.singleton(new Description("8635753033", 1, true, "900000000000012033", "3311481044",
+																									"en", "900000000000013044", "Test term", "900000000000448022"))).build();
+		traceabilityLogService.logActivity(null, commit, persistedComponents, false, "Delta");
+		activitiesLogged.stream().map(activity -> activity.getCommitComment().contains("RF2 Import - Updating concept Test FSN")).forEach(Assert::assertTrue);
```
kaicode, line 182: Nice assertion including the new message prefix.
Be careful with expectations on collections, if no activities are logged this will still pass..

kaicode (changes requested):
I find the set of classes in the new `persistedcomponent` package overkill for this one-off and context specific task. Please consider refactoring into a single helper class. Using generics in the internal method signature rather than at class level could solve the issue in the same way?
Loading all component changes from the store from a specific commit is a very specific thing to do and may only ever be done by the traceability service. Other services either already have the objects in memory or need a specific set of objects, this is already catered for using the versionControlHelper. Feel free to call if you want to talk it though more : )

src/main/java/org/snomed/snowstorm/core/data/services/TraceabilityLogService.java
Tyler-Cooksey: Fixed

src/main/java/org/snomed/snowstorm/core/data/services/TraceabilityLogService.java
Tyler-Cooksey: Fixed

src/main/java/org/snomed/snowstorm/core/rf2/rf2import/ImportService.java
Tyler-Cooksey: Fixed

src/test/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceTest.java
Tyler-Cooksey: Fixed

src/main/java/org/snomed/snowstorm/core/data/services/TraceabilityLogService.java
```diff
@@ -65,18 +74,69 @@ public TraceabilityLogService() {
 	}
 
 	@Override
-	public void preCommitCompletion(Commit commit) throws IllegalStateException {
-		if (commit.getCommitType() != CONTENT) {// Only use for rebase / promotion
-			logActivity(SecurityUtil.getUsername(), commit, new PersistedComponents());
+	public void preCommitCompletion(final Commit commit) throws IllegalStateException {
+		// Most commits don't use this because ConceptService calls logActivity directly.
```
kaicode, line 78: very concise 👍

src/main/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceHelper.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.stereotype.Service;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import static io.kaicode.elasticvc.api.VersionControlHelper.LARGE_PAGE;
+
+@Service
+public class TraceabilityLogServiceHelper {
```
kaicode, line 21: Great improvement. Same functionality with generics in a single class 👍

src/main/java/org/snomed/snowstorm/core/rf2/RF2Constants.java
```diff
@@ -2,11 +2,15 @@
 
 import java.util.regex.Pattern;
 
-public class RF2Constants {
+public final class RF2Constants {
```
kaicode, line 5: Nice idea.

kaicode (approved):
Looks good to me!
Feel free to mege when you are ready.

PR #192, opened 2020-12-04
Fix MAINT-1322, Rewrite of historical association merge.

kaicode:
Merging of the historical associations of a concept or description can happen either by saving a new version of a concept through the API or by merging two branches together. When merging two branches we were finding that duplicate records were created because two refset members with the same refset id and association target existed on both sides of the merge. We try to favour the released and active records so a comparator is now used for this.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
@@ -361,21 +361,26 @@ public class ConceptUpdateHelper extends ComponentService {
 
 	private ReferenceSetMember getBestRefsetMember(String refsetId, String additionalFieldKey, String additionalFieldValue, Set<ReferenceSetMember> newVersionMembers, Set<ReferenceSetMember> existingVersionMembers) {
 		ReferenceSetMember bestMember = null;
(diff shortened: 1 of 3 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -296,59 +297,94 @@ private void validateConcepts(Collection<Concept> concepts) {
...
+			String associationRefsetId = Concepts.historicalAssociationNames.inverse().get(associationTypeName);
+			for (String associationValue : newVersionAssociations.get(associationTypeName)) {
+
+				ReferenceSetMember bestMember = getBestRefsetMember(associationRefsetId, ReferenceSetMember.AssociationFields.TARGET_COMP_ID, associationValue, newVersionMembers, existingVersionMembers);
+				if (bestMember != null) {
+					// Keep
+					membersToKeep.add(bestMember);
+				} else {
+					// Create new
+					bestMember = new ReferenceSetMember(newComponentVersion.getModuleId(), associationRefsetId, newComponentVersion.getId());
```
Tyler-Cooksey, line 326: Calling this constructor sets the active state to true implicitly already

Tyler-Cooksey (approved):

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
kaicode: Yes, good point.

PR #194, opened 2020-12-09
CDI-90 -- Extract concrete data type from snomed-owl-toolkit / mrcm

dmcgihtsdo:
CDI-90 is concerned with extracting the data type of a concrete value from both the snomed-owl-toolkit and the MRCM. 

To test: 

* Request a Relationship with concrete data (the concrete dta type will be retrieved from mrcm)
* Request a Concept with concrete data (the concrete data type will be retrieved from snomed-owl-toolkit)

Note: This code is related to snomed-owl-toolkit@feature/CDI-90.


src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
```diff
@@ -187,30 +187,28 @@ public boolean isGrouped () {
...
 	}
 
 	public void setConcreteValue(ConcreteValue concreteValue) {
 		this.concreteValue = concreteValue;
 	}
 
-	public void setConcreteValue(String value) {
-		if (value != null) {
-			this.concreteValue = new ConcreteValue(value);
+	public void setConcreteValueFromExternal(org.snomed.otf.owltoolkit.domain.Relationship.ConcreteValue concreteValue) {
```
dmcgihtsdo, line 197: I changed the name of the method to setConcreteValueFromExternal as there would have been two methods named setConcreteValue that had a parameter of ConcreteValue.

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
```diff
@@ -187,30 +187,28 @@ public boolean isGrouped () {
...
+			this.destinationId = null;
+			this.target = null;
 		}
-
-		this.destinationId = null;
-		this.value = value;
-		this.target = null;
 	}
 
 	public boolean isConcrete() {
```
dmcgihtsdo, line 206: The isConcrete method was situation dependent, so I changed the implementation to try and cater for the situations.

src/main/java/org/snomed/snowstorm/core/data/services/ServiceError.java
```diff
@@ -0,0 +1,11 @@
+package org.snomed.snowstorm.core.data.services;
```
dmcgihtsdo, line 1: I created ServiceError so I could throw it In RelationshipService if there is a problem when reading MRCM data (line 95).

src/main/java/org/snomed/snowstorm/mrcm/MRCMLoader.java
```diff
@@ -1,6 +1,9 @@
 package org.snomed.snowstorm.mrcm;
```
dmcgihtsdo, line 1: In MRCMLoader, I have moved the bulk of the logic into private methods. As well as this, I am storing the MRCM object in a map. Subsequent requests will use the MRCM object from the map if the same branchPath is used.

src/test/java/org/snomed/snowstorm/core/data/domain/ConcreteValueTest.java
```diff
@@ -1,136 +1,363 @@
 package org.snomed.snowstorm.core.data.domain;
```
dmcgihtsdo, line 1: The ConcreteValue class changed a lot; tests reflect this.

src/main/java/org/snomed/snowstorm/core/data/services/ServiceError.java
dmcgihtsdo: I've deleted this and using RuntimeServiceException instead.

src/main/java/org/snomed/snowstorm/mrcm/MRCMLoader.java
```diff
@@ -18,116 +21,159 @@
...
+    @Autowired
+    private ReferenceSetMemberService memberService;
+
+    @Override
+    public void preCommitCompletion(final Commit commit) throws IllegalStateException {
+        if (!this.cache.isEmpty()) {
+            final Branch branch = commit.getBranch();
+            final String branchPath = branch.getPath();
+
+            this.cache.remove(branchPath);
```
CoderMChu, line 48: It is good that the cache got cleared after a commit is made on a branch to prevent memory leak.

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
```diff
@@ -187,30 +187,28 @@ public boolean isGrouped () {
...
-		return this.value != null && this.destinationId == null;
+		final boolean hasValue = this.value != null;
+		final boolean hasConcreteValue = this.concreteValue != null;
+		final boolean hasNoDestinationId = this.destinationId == null || this.destinationId.isEmpty();
+
+		return (hasValue && hasNoDestinationId) || (hasConcreteValue && hasNoDestinationId);
 	}
 
 	@Override
```
CoderMChu, line 214: I think it might be clearer to set destinationId to null in the default constructor and use this.value != null && this.destinationId == null

src/main/java/org/snomed/snowstorm/core/data/services/RelationshipService.java
```diff
@@ -46,9 +52,49 @@
...
+
+				for (AttributeRange attributeRange : attributeRanges) {
+					final String referencedComponentId = attributeRange.getReferencedComponentId();
+					if (typeId.equals(referencedComponentId)) {
+						final String rangeConstraint = attributeRange.getRangeConstraint();
+						final String shorthandDataType = rangeConstraint.substring(0, 3);
+
+						relationship.setConcreteValue(
+								ConcreteValue.fromShorthand(value, shorthandDataType)
+						);
```
CoderMChu, line 90: There might be multiple attribute ranges for a given attribute id as the range is 1-to-1 mapping to a attribute domain.

src/main/java/org/snomed/snowstorm/core/data/services/ServiceError.java
CoderMChu: It might be better to wrap the ServiceException.
throw new ServiceError("Trouble loading active MRCM data", e);

src/main/java/org/snomed/snowstorm/core/data/services/AxiomConversionService.java
```diff
@@ -115,11 +115,7 @@ private AxiomRelationshipConversionService setupConversionService(String branchP
 					internalRelationship.setSourceId(sourceId != null ? sourceId.toString() : null);
 					internalRelationship.setGroupId(externalRelationship.getGroup());
 					internalRelationship.setInferred(false);
-
-					final org.snomed.otf.owltoolkit.domain.Relationship.ConcreteValue externalConcreteValue = externalRelationship.getValue();
-					if (externalConcreteValue != null) {
-						internalRelationship.setConcreteValue(externalConcreteValue.getRF2Value());
-					}
+					internalRelationship.setConcreteValueFromExternal(externalRelationship.getValue());
```
CoderMChu, line 118: To avoid having two ConcreteValues in the Relationship class you could have a method like this setConcreteValue(String type, String value).

CoderMChu (changes requested):
@dmcgihtsdo It looks good. It just needs some minor tweaks.

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
dmcgihtsdo: This method call has now been reverted back to setConcreteValue with two String arguments.

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
dmcgihtsdo: Like @CoderMChu suggested, the destinationId is now set to null in the constructor; the implementation of isConcrete is now much simpler.

src/main/java/org/snomed/snowstorm/core/data/services/ServiceError.java
dmcgihtsdo: Thanks for the suggestion - I have added that to the RuntimeServiceException.

src/main/java/org/snomed/snowstorm/core/data/domain/Relationship.java
dmcgihtsdo: I have made that change @CoderMChu . When I made the change, I was getting a NPE whenever requesting Relationships; so I have added a null check to the expandSourceTypeAndDestination method in RelationshipController.

src/main/java/org/snomed/snowstorm/core/data/services/RelationshipService.java
dmcgihtsdo: I have added a break statement to exit the for-loop.

src/main/java/org/snomed/snowstorm/core/data/services/AxiomConversionService.java
dmcgihtsdo: I have made those changes.

src/main/java/org/snomed/snowstorm/rest/RelationshipController.java
```diff
@@ -83,9 +83,12 @@
 	private void expandSourceTypeAndDestination(String branch, List<Relationship> relationships, List<LanguageDialect> languageDialects) {
```
dmcgihtsdo, line 83: I have changed the default constructor for Relationship to set the id to null instead of a blank String. Because of this, Elasticsearch was throwing an exception because of null fields; to fix this I have added a null check to RelationshipController.

CoderMChu: @dmcgihtsdo  Good catch. Can you just add a space in the if statement below before '(' and after ')' please?

dmcgihtsdo: Fixed. My bad, I normally format before pushing.

CoderMChu (approved):
It looks good now. Thanks @dmcgihtsdo

src/main/java/org/snomed/snowstorm/mrcm/MRCMLoader.java
kaicode, line 48: Try to strike a balance between declaring variables and just calling getters to provide method parameters in a single line. Declaring variable can make code more readable if there is lots going on and you are trying to make a point and draw attention. This is usually when a variable will be used more than once. 
In this case I don't think any variables are needed, this would be fine `this.cache.remove(commit. getBranch(). getPath())`? In fact that could be the only line of this method, if block does not add anything : )

src/main/java/org/snomed/snowstorm/core/data/services/RelationshipService.java
```diff
@@ -46,9 +52,49 @@
 	@Autowired
 	private ConceptUpdateHelper conceptUpdateHelper;
 
+	@Autowired
+	private MRCMLoader mrcmLoader;
+
 	public Relationship findRelationship(String branchPath, String relationshipId) {
```
kaicode, line 58: This method will be used when fetching a single relationship so this API endpoint probably works: `GET /{branch}/relationships/{relationshipId}`
We also need the concrete types set when viewing the inferred relationships within a concept using `GET /browser/{branch}/concepts/{conceptId}` and viewing the list under `{"relationships" : [] }`. 
The setConcreteValueFromMRCM method may have to become protected or public and allow a collection of relationships to be passed in. I think it would be good separation of concerns for the MRCMLoader to have a method like getConcreteType(attributeType) so the relationship service is doing less work. The data type is an MRCM concern really. The AttributeRange could have a field for the data type which gets populated on load.

src/main/java/org/snomed/snowstorm/mrcm/MRCMLoader.java
```diff
@@ -18,116 +21,159 @@
...
+        List<AttributeRange> attributeRanges = new ArrayList<>();
+        for (ReferenceSetMember member : attributeRangeMembers) {
+            // id	effectiveTime	active	moduleId	refsetId	referencedComponentId	rangeConstraint	attributeRule	ruleStrengthId	contentTypeId
+            if (member.isActive()) {
+                attributeRanges.add(new AttributeRange(
+                        member.getMemberId(),
+                        member.getEffectiveTime(),
+                        member.isActive(),
+                        member.getReferencedComponentId(),
+                        member.getAdditionalField("rangeConstraint"),
```
kaicode, line 147: Could set the concrete type here too, or get the constructor method to set it, if there is one.

kaicode (changes requested):
Really good progress on CDI-90. One gap identified and some refactoring suggested. Thanks @dmcgihtsdo

src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
```diff
@@ -19,41 +19,133 @@
     private final String value;
 
     public enum DataType {
-        DECIMAL("DECIMAL"),
-        INTEGER("INTEGER"),
-        STRING("STRING");
+        DECIMAL("DECIMAL", "dec"),
```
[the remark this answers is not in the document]
dmcgihtsdo: Inline with your comment on the snomed-owl-toolkit pull request, I have removed the name field from the enum.

src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
```diff
@@ -19,41 +19,133 @@
...
-        final boolean isConcreteNumber = value.startsWith("#");
-        final boolean isConcreteString = value.startsWith("\"");
+        /**
+         * Create instance of DataType.
+         *
+         * @param name Data type for concrete value.
+         * @return Instance of DataType.
+         * @throws IllegalArgumentException If name is not recognised.
+         */
+        public static DataType fromName(final String name) {
```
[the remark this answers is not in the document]
dmcgihtsdo: I have removed the method.

src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
```diff
@@ -19,41 +19,133 @@
...
+         * @param dataType Shorthand data type for concrete value.
+         * @return Instance of DataType.
+         * @throws IllegalArgumentException If name is not recognised.
+         */
+        public static DataType fromShorthand(final String dataType) {
+            if (dataType == null) {
+                throw new IllegalArgumentException("'dataType' cannot be null.");
+            }
+
+            switch (dataType.toLowerCase()) {
```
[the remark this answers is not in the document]
dmcgihtsdo: Great idea - thanks for the suggestion.

src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
```diff
@@ -19,41 +19,133 @@
...
+     * @return Instance of ConcreteValue.
+     * @throws IllegalArgumentException If value does not have concrete prefix.
+     */
+    public static ConcreteValue fromShorthand(String value,
+                                              final String dataType) {
+        if (value == null || dataType == null) {
+            throw new IllegalArgumentException();
+        }
+
+        if (value.startsWith("#")) {
```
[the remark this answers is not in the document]
dmcgihtsdo: I'm not entirely sure what you mean @kaicode - the method would have failed if the prefix wasn't present. I have, however, updated the methods to trust their arguments and presume the format of them etc.

src/main/java/org/snomed/snowstorm/core/data/services/RelationshipService.java
```diff
@@ -46,9 +52,49 @@
...
+		}
+
+		final Relationship relationship = relationships.getContent().get(0);
+		setConcreteValueFromMRCM(relationship, branchPath);
+
+		return relationship;
+	}
+
+	private void setConcreteValueFromMRCM(final Relationship relationship,
+										  final String branchPath) {
```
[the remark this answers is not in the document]
dmcgihtsdo: Thanks for sharing! I never thought about other platforms and IDEs - so that is good to know. All methods should now have their arguments inline.

src/main/java/org/snomed/snowstorm/mrcm/MRCMLoader.java
dmcgihtsdo: Good idea - I have added an additional constructor to AttributeRange with a ConcreteValue.DataType field option.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -99,6 +99,9 @@
 	@Autowired
 	private ConceptAttributeSortHelper conceptAttributeSortHelper;
 
+	@Autowired
```
dmcgihtsdo, line 102: I have updated the doFind method in ConceptService to set the ConcreteValue of Relationships.

src/main/java/org/snomed/snowstorm/core/data/services/RelationshipService.java
dmcgihtsdo: Thank you for picking up on that @kaicode. 

I have implemented both suggestions.

src/main/java/org/snomed/snowstorm/core/data/domain/ConcreteValue.java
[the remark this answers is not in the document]
kaicode: Good point, I see you throw an exception if the hash and quote are missing.

src/main/java/org/snomed/snowstorm/core/data/services/RelationshipService.java
```diff
@@ -46,9 +50,56 @@
...
+
+	/**
+	 * Set the ConcreteValue of each Relationship.
+	 *
+	 * @param branchPath    The branchPath to load active MRCM data from.
+	 * @param relationships The Relationships to update.
+	 * @throws RuntimeServiceException When there is an issue reading MRCM.
+	 */
+	public void setConcreteValueFromMRCM(String branchPath, Relationship... relationships) {
+		final BranchCriteria branchCriteria = versionControlHelper.getBranchCriteria(branchPath);
```
kaicode, line 76: This method is cheap-ish but not free. We will be creating a new branch criteria for every inferred relationship when loading a concept even if it's not used because the MRCM is cached. It would be better to just pass the branch path to the MRCM loader because usually the MRCM will be cached and the branch criteria will not be needed. The mrcm loader will need access to the versionControlHelper so it can create a BranchCriteria if needed.

kaicode (changes requested):
Suggested a small but important optimisation, otherwise you are there!

src/main/java/org/snomed/snowstorm/mrcm/MRCMLoader.java
dmcgihtsdo: I have added a new method to MRCMLoader; this new method is the method that has the cache. The older method remains the same.

kaicode (approved):
Looking good!

Issue #179, opened 2020-11-10
Add the spring boot actuator dependency


manuel-palacio:
https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html

Would make snowstorm easier to monitor by adding at least the /actuator/health endpoint

kaicode:
Added SI Jira ticket `MAINT-1523`

kaicode:
GitHub is not allowing my to assign @jbarcas to this ticket but he has made the change and done the testing 😄 .

kaicode:
This is included in the [6.0.3 release](https://github.com/IHTSDO/snowstorm/releases/tag/6.0.3), thanks to @jbarcas 🙇

PR #200, opened 2021-01-06
#179 Add Spring Boot Actuator dependency

jbarcas:


kaicode:
Hi @jbarcas, thanks for this pull request. Could you provide a screenshot of this working please? Could you also list the functional changes that enabling the Spring Boot Actuator gives and if there is any security considerations like exposing information about the machine or any config? I've not picked this up yet because I have not had time to read the docs. Thank you.

kaicode:
I've just seen your email that has some of this info but please quote it here for the record. thank you.

jbarcas:
Sure!

Here you are the screenshot:

image

Regarding the security issues: The built-in endpoints the actuator provides must be both _enabled_ and _exposed_ to be available. All endpoints are _enabled_ by default, but just `info` and `health` are _exposed_ via http (see [exposed expoints](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-endpoints-exposing-endpoints)) by default. Also, it's possible to _disable_ all endpoints by default via properties and _enable_ particular ones (disable an endpoint means to entirely remove it from the application context). Besides, Spring Security secures all endpoints by default using its content-negotiation strategy. It depends on which endpoints you want to expose but, IMO, there are no security concerns with the default config.

Source: [Spring docs](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html).

Cheers.

kaicode (approved):
Looks good.
Thank you for doing the reading and providing a quick summary of the changes for this feature. Since only the `info` and `health` endpoints are enabled by defaults I have no concerns.

PR #203, opened 2021-01-11
MAINT-1353 - Added support for POST operation on ValueSet Expansion.

Tyler-Cooksey:
The reason for : `@ResourceParam String rawBody` was because there isn't a nice way to encapsulate the POST body which is sent up so after doing some research, this is a work around which under a POST operation, contains the body then a simple cast can occur that maps it into the `Parameters` object. Under the GET operation, this is not used and the String content is empty so it's only ever populated at runtime when a POST operation occurs.

Currently writing unit tests.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/fhir/exceptions/MissingParameterException.java
+++ b/src/main/java/org/snomed/snowstorm/fhir/exceptions/MissingParameterException.java
@@ -1,0 +1,8 @@
+package org.snomed.snowstorm.fhir.exceptions;
+
+public class MissingParameterException extends RuntimeException {
+
+	public MissingParameterException(final String message) {
+		super(message);
+	}
+}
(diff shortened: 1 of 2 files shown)


pgwilliams:
Hi @Tyler-Cooksey  - looks gorgeous thank you.   Obviously rawBody is the vital addition which does the business and I look forward to trying that out.   Some more general comments:
1.  I think it would be good for your pull request to include all the code changes AND the unit tests AND the documentation updates.   I'd be very happy to review a commit on some branch if you'd like some "going in the right direction" feedback before committing additional time.   Ideally the unit tests would be reviewed in conjunction with the code updates.
2. Don't be afraid of code comments, especially if you've figured something out in the process of writing the code.   When it comes to FHIR, we're all still figuring it out and your comments might help people even outside of this organisation.
3. It was "my bad" that all those parameters were added to the expand call in a long list.   It was growing somewhat organically at the time.   But now that you're looking at it with fresh eyes, I think it would be a good time to wrap that up in some Parameters object so our method call can be simplified.    You'll then need to code like   p.getForceSystemVersion() rather than just having the variable already present.

Great work though, I'll merge now.

Issue #204, opened 2021-01-11
Badly-generated API definition


evergreen-lee-campbell:
The JSON specification for the API makes reference to a "Set" definition, which is not present in `#/definitions` (example in excerpt, below), meaning Client SDK generators do not produce compileable TypeScript (and likely other languages in which `Set` is a primitive or otherwise common type.

```json
		"Description": {
			"type": "object",
			"properties": {
				"acceptabilityMap": {
					"type": "object",
					"additionalProperties": {
						"type": "string"
					}
				},
				"acceptabilityMapFromLangRefsetMembers": {
					"type": "object",
					"additionalProperties": {
						"type": "string"
					}
				},
				"active": {
					"type": "boolean"
				},
				"associationTargets": {
					"type": "object",
					"additionalProperties": {
						"$ref": "#/definitions/Set"
					}
				},
				...
```

evergreen-lee-campbell:
I've done some digging, and this is actually a problem with Springfox. I'm currently investigating whether a simple version upgrade does the trick...

evergreen-lee-campbell:
It does, but there are some breaking changes. I've raised the following pull request: https://github.com/IHTSDO/snowstorm/pull/205

kaicode:
Closing as the PR was tested for some time but finally we chose not to pursue the upgrade. Further explanation on the PR.

PR #205, opened 2021-01-12
#204 - Upgraded to Springfox 3.0.0 to resolve Map definitions issue in api-docs

evergreen-lee-campbell:


kaicode:
Testing this locally.

kaicode:
@evergreen-lee-campbell thanks for getting this started. 

Unfortunately I'm not able to view the generated swagger page. The root servlet at `http://localhost:8080/` still redirects to `http://localhost:8080/swagger-ui.html` but there is nothing there. I also tried `http://localhost:8080/swagger-ui/` and `http://localhost:8080/swagger-ui/index.html` but nothing loaded. Is there another URL I should use?

I found that adding a dependency for `slf4j-api` cleared the errors at startup but still no luck.

evergreen-lee-campbell:
Didn't spot those startup errors, sorry. I've included the same package and they've gone, and updated the "swagger-ui" URL too.

evergreen-lee-campbell:
I am finally round to looking at this again after a bit of a hiatus. It appears the LGTM analysis is failing, but I can't see the logs to understand why. It wasn't failing when my branch was (erroneously) trying to pull back into master, but is now failing that I'm pointing to develop. Is the problem, therefore with develop? I've also added some tests to cover some of the change I've made. Admittedly, the test on "SecurityAndUriConfig" is a bit of a token effort to get a "green tick" from the coverage checks you've implemented. 

What else, if anything, would you like from me in order to get this into a release?

kaicode:
Hi @evergreen-lee-campbell, thank you for your efforts trying out Springfox 3.

If you use Swagger a lot like my team do the "Try it out" button in Swagger 3 is quite annoying! I can see others think the same (Swagger issues [#2806](https://github.com/swagger-api/swagger-ui/issues/2806) and [#6528](https://github.com/swagger-api/swagger-ui/issues/)).

We will try this out in our development environments and see how we get on...

kaicode:
We have been trying out Swagger 3.0.0 (using the commits from this PR) in our dev environment for the last week.

We have had a mixed experience. It's nice to be compliant with the latest OpenAPI standard but the new Swagger UI leaves a lot to be desired...

These are a few of the issues we found with the new UI:
- The "Try it out" button has become more annoying the more we use it.
- The request objects for some endpoints have not been populated properly, where they were on the previous version. For example the create code systems endpoint has a blank request model.
- The RF2 import file operation will not work. The file upload field is rendered as a text area in the Swagger UI.
- The endpoint parameters are sorted alphabetically rather than preserving the intuitive order use in the Java controller class. For example in the concept search endpoint the pagination options are scattered through the rest of the parameters rather than placed at the end. This makes the API less intuitive to use.

While I appreciate the work put in here and we understand that Swagger 3 gives better code stub generation unfortunately I am going to have to turn down this PR because the Swagger 3 UI is a big step backwards and will have a negative impact for most.

I hope you can understand my position @evergreen-lee-campbell. I have not taken this decision lightly but I have to act in the best interest of the majority of users and I think people use the Swagger UI more than they generate code stubs.

I hope you will not be put off raising PRs in the future, we really appreciate contributions to make this project better.

Kind regards,
Kai Kewley

evergreen-lee-campbell:
Hi Kai,

I totally understand. Thanks for looking into it. I'll keep my fork maintained until the Swagger 3 UI improves, I guess.

I'm not at all put off! I love what you've done with this project, and if I encounter any other ways to contribute, I will indeed. :)

All the best,
Lee

PR #206, opened 2021-01-12
MAINT-1353 - Added support for POST operation on ValueSet Expansion, refactored code to use a builder and added an example to the README for a POST operation.

Tyler-Cooksey:
New and improved PR with code refactor and unit tests.

### DIFF
--- a/docs/fhir-resources/valueset-expansion.md
+++ b/docs/fhir-resources/valueset-expansion.md
@@ -3,43 +3,6 @@
 ### Implicit ValueSets (ie intensionally defined). 
 See  [https://www.hl7.org/fhir/snomedct.html#implicit]
 
-## POST Method Request Example(s)
-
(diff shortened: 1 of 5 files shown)


pgwilliams:
Thanks @Tyler-Cooksey I love the builder pattern.   But the purpose of creating that object (apologies for not being clearer about this) was to simplify the call to expand method so that you bundle this loose collection of things up into a single object and pass _that_, rather than putting all these strings into a container and then immediately unpacking them again.  The intention was to fix an embarrassingly (my embarrassment) long method signature.

Also in your documentation, could you move the POST example to the bottom of the file (since it's quite long) and provide it as curl so that it could be immediately cut and pasted to try out on the command line?   Thanks!   You could condense your brackets onto the same lines as the elements/values to save space.

Tyler-Cooksey:
Sure no problem, will cleanup the code more now.

PR #196, opened 2020-12-17
BROWSE-241 -- View history of a Concept

dmcgihtsdo:
BROWSE-241 is concerned with updating Snowstorm's API to include a method for returning the history of a Concept.

PR #211, opened 2021-01-19
BROWSE-241 -- View history of a Concept

dmcgihtsdo:
BROWSE-241
-----
BROWSE-241 is concerned with updating Snowstorm's API to include a method for returning the history of a Concept.

This PR is an extension to (https://github.com/IHTSDO/snowstorm/pull/196) as it changes the query used for requesting a Concept's history.


src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -58,6 +61,8 @@
 @Service
 public class ConceptService extends ComponentService {
 	private static final Map<ComponentType, Class<? extends DomainEntity<?>>> COMPONENT_DOCUMENT_TYPES = new HashMap<>();
+	private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyyMMdd")
```
dmcgihtsdo, line 64: @kaicode , I'm always wary when it comes to using time in Java. Are you happy with my approach of using `DateTimeFormatter` and `Instant` ?

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -153,55 +158,62 @@ public Concept find(String id, List<LanguageDialect> languageDialects, BranchTim
 	}
 
 	public ConceptHistory loadConceptHistory(String conceptId, List<CodeSystemVersion> codeSystemVersions, boolean showFutureVersions) {
+		int now = Integer.parseInt(DATE_TIME_FORMATTER.format(Instant.now()));
 		Map<String, BranchCriteria> branchCriteria = new HashMap<>();
 		for (CodeSystemVersion codeSystemVersion : codeSystemVersions) {
-			String branchPath = codeSystemVersion.getBranchPath();
-			branchCriteria.put(branchPath, versionControlHelper.getBranchCriteria(branchPath));
+			if (showFutureVersions || codeSystemVersion.getEffectiveDate() < now) {
```
dmcgihtsdo, line 164: As suggested, I am now only adding a branchCriteria to my map if showFutureVersions is indeed true or if the codeSystemVersion is not in the future; this will prevent the logic from being carried out directly on ES.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -153,55 +158,62 @@ public Concept find(String id, List<LanguageDialect> languageDialects, BranchTim
...
+				String branchPath = codeSystemVersion.getBranchPath();
+				branchCriteria.put(branchPath, versionControlHelper.getBranchCriteria(branchPath));
+			}
 		}
 
-		Function<String, BoolQueryBuilder> defaultBoolQueryFunction = cId -> {
-			BoolQueryBuilder someReleaseBranch = boolQuery();
-			BoolQueryBuilder boolQueryBuilder = boolQuery();
+		BiFunction<String, ComponentType, BoolQueryBuilder> defaultFullQuery = (cId, cT) -> {
```
dmcgihtsdo, line 170: I have renamed the variables associated with `boolQuery` to help readability and nested where possible.

src/test/java/org/snomed/snowstorm/rest/ConceptControllerTest.java
```diff
@@ -662,7 +662,6 @@ public void findConceptHistory_ShouldReturnExpectedConceptHistory_WhenShowFuture
 		});
 		ConceptHistory conceptHistory = responseEntity.getBody();
 		List<ConceptHistory.ConceptHistoryItem> history = conceptHistory.getHistory();
-		ConceptHistory.ConceptHistoryItem januaryRelease = conceptHistory.getConceptHistoryItem("20200131").get();
```
dmcgihtsdo, line 665: This line was unused.

src/test/java/org/snomed/snowstorm/rest/ConceptControllerTest.java
```diff
@@ -750,7 +749,6 @@ public void findConceptHistory_ShouldReturnExpectedConceptHistory_WhenShowFuture
...
```
dmcgihtsdo, line 753: This line was unused.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
kaicode: Looks good to me. I still tend to use `SimpleDateFormat` but it's not thread safe. This Java 8 method looks better 👍

src/main/java/org/snomed/snowstorm/core/data/domain/Concept.java
```diff
@@ -29,6 +29,7 @@
 	public interface Fields extends SnomedComponent.Fields {
 		String CONCEPT_ID = "conceptId";
 		String DEFINITION_STATUS_ID = "definitionStatusId";
+		String PATH = "path";
```
kaicode, line 32: This could go in the `SnomedComponent` class.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
kaicode: The call to `codeSystemService.findAllVersions(codeSystem.getShortName(), showFutureVersions)` in the concept controller should already take care of the filtering. You shouldn't need to check the version effectiveDate here. You may not need `showFutureVersions` as a parameter of this method (`loadConceptHistory`).

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -153,55 +158,62 @@ public Concept find(String id, List<LanguageDialect> languageDialects, BranchTim
...
+			BoolQueryBuilder fullQuery = boolQuery();
+			fullQuery.should(
```
kaicode, line 172: This should be a `must` ?

src/test/java/org/snomed/snowstorm/rest/ConceptControllerTest.java
kaicode: On line 498 the componentType could be `Concept` every time.

kaicode (commented):
Looks good. I couple of minor suggestions made.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
dmcgihtsdo: Fixed.

src/test/java/org/snomed/snowstorm/rest/ConceptControllerTest.java
dmcgihtsdo: @kaicode , I'm sorry I don't understand. January has 4 component types (Concept, Description, Relationship & Axiom) which is asserted in the for-loop and so it's not always Concept?

src/main/java/org/snomed/snowstorm/core/data/domain/Concept.java
dmcgihtsdo: Good idea.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
dmcgihtsdo: Yep, good catch. Changed it and re-tested; all working as expected.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
dmcgihtsdo: I have removed this as the `codeSystemService.findAllVersions()` method call does indeed return the correct versions.

src/test/java/org/snomed/snowstorm/rest/ConceptControllerTest.java
kaicode: The key word is _could_. If they were all Concept, or two Description and two Relationship the assertion would still pass right?

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
kaicode: Looks good. I find it easier to read when a separate boolQuery is used to hold the should clauses rather than holding should and must clauses against the same boolQuery but the `minimumShouldMatch` method seems to work. This is fine.

dmcgihtsdo: Ah, I also had that idea and originally had that code as it was "grouped" better. I figured it was OTT, but good to know going forward!

src/test/java/org/snomed/snowstorm/rest/ConceptControllerTest.java
dmcgihtsdo: Thanks, I get you now; I have fixed this by being more explicit in my tests. I encountered a minor gotcha in that my response was being deserialised into a HashSet rather than my defined TreeSet which was of course useless for my tests/assertions. To get around this I added the `@JsonDeserialize` annotation to `ConceptHistory`.

kaicode (approved):
Sorted.

PR #212, opened 2021-01-20
CDI-45 Concrete Domains Drools Snowstorm Implementation

QuyenLy87:
Drools Snowstorm Implementation against Concrete domain value

### DIFF
--- a/src/main/java/org/snomed/snowstorm/validation/domain/DroolsRelationship.java
+++ b/src/main/java/org/snomed/snowstorm/validation/domain/DroolsRelationship.java
@@ -54,7 +54,7 @@ public class DroolsRelationship implements org.ihtsdo.drools.domain.Relationship
 
 	@Override
 	public String getConcreteValue() {
-		return null;
+		return relationship.getValue();
 	}
 
 	@Override
--- a/src/test/java/org/snomed/snowstorm/validation/DroolValidationServiceTest.java
+++ b/src/test/java/org/snomed/snowstorm/validation/DroolValidationServiceTest.java
@@ -1,0 +1,192 @@
+package org.snomed.snowstorm.validation;
+
+import io.kaicode.elasticvc.api.BranchService;
+import org.apache.commons.io.FileUtils;
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.lib.TextProgressMonitor;
+import org.ihtsdo.drools.response.InvalidContent;
+import org.ihtsdo.drools.response.Severity;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.snomed.snowstorm.AbstractTest;
+import org.snomed.snowstorm.TestConfig;
+import org.snomed.snowstorm.core.data.domain.*;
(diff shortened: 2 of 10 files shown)


src/test/java/org/snomed/snowstorm/validation/DroolValidationServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+        index++;
+        assertEquals(Severity.WARNING, invalidContents.get(index).getSeverity());
+        assertEquals("Active FSN should end with a valid semantic tag.", invalidContents.get(index).getMessage());
+
+        index++;
+        assertEquals(Severity.WARNING, invalidContents.get(index).getSeverity());
+        assertEquals("A concept's semantic tags should be compatible with those of the active parents.", invalidContents.get(index).getMessage());
+    }
+
+    private void checkoutDroolsRules() throws GitAPIException, IOException {
```
kaicode, line 169: Very cool that you can check out another project using Java, however I have concerns with this approach. I would rather keep the repositories separate otherwise the Snowstorm build could fail when the content in the drools rules project changes.
Could you remove this git clone functionality and just clone two or three snomed drools rules into the snowstorm test resources instead please?
You could copy the rule you are using for FSN validation and a rule to validate relationships to check that doesn't break with concrete domains.

src/test/java/org/snomed/snowstorm/validation/DroolValidationServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @Test
+    void testValidateConceptAgainstConcreteValue() throws ServiceException {
+        Concept foundConcept = conceptService.find("100001", DEFAULT_BRANCH);
+        assertNotNull(foundConcept);
+
+        foundConcept.getDescriptions().stream().forEach(description -> description.setAcceptabilityMap(Collections.singletonMap(Concepts.US_EN_LANG_REFSET,
+                Concepts.descriptionAcceptabilityNames.get(PREFERRED))));
+
+        foundConcept.addRelationship(new Relationship("100002", ISA, SNOMEDCT_ROOT).setCharacteristicTypeId(Concepts.STATED_RELATIONSHIP));
```
kaicode, line 145: We no longer use stated relationships although the Snowstorm DroolValidationService will convert the fake "relationships" within axioms into stated DroolsRelationship objects for validation. 
In order to use this stuff in the test please:
- change this one to an inferred relationship
- also add an axiom to the concept with a relationship within that
Thanks!

src/main/java/org/snomed/snowstorm/validation/domain/DroolsRelationship.java
```diff
@@ -54,7 +54,7 @@ public String getCharacteristicTypeId() {
 
 	@Override
 	public String getConcreteValue() {
-		return null;
+		return relationship.getValue();
```
kaicode, line 57: This looks good. Nice simple change 😄

kaicode (changes requested):
Thanks for writing these unit tests, things are moving in the right direction. Please make the requested changes and let me know.
Cheers

src/test/java/org/snomed/snowstorm/validation/DroolValidationServiceTest.java
QuyenLy87: The unit tests have been updated @kaicode

src/test/java/org/snomed/snowstorm/validation/DroolValidationServiceTest.java
QuyenLy87: All above-changed requests have been fixed. @kaicode

kaicode (approved):
Looks good!

PR #214, opened 2021-01-21
MAINT-1462 - Faster concept save by combining save and validate functions

Tyler-Cooksey:


src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -369,30 +413,45 @@ public ConceptReferencesResult findConceptReferences(
...
 
-	@RequestMapping(value = "/browser/{branch}/concepts/{conceptId}", method = RequestMethod.PUT)
+	@PutMapping(value = "/browser/{branch}/concepts/{conceptId}/{validate}")
 	@PreAuthorize("hasPermission('AUTHOR', #branch)")
 	@JsonView(value = View.Component.class)
-	public ConceptView updateConcept(
+	public ResponseEntity<ConceptView> updateConcept(
 			@PathVariable String branch,
 			@PathVariable String conceptId,
+			@PathVariable(required = false) Boolean validate,
```
kaicode, line 441: Same as above, please use request parameter.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -369,30 +413,45 @@ public ConceptReferencesResult findConceptReferences(
 		return new ConceptReferencesResult(typeSets, conceptReferencesPage.getPageable(), conceptReferencesPage.getTotalElements());
 	}
 
-	@RequestMapping(value = "/browser/{branch}/concepts", method = RequestMethod.POST)
+	@PostMapping(value = "/browser/{branch}/concepts/{validate}")
```
kaicode, line 416: I don't mind you changing `@RequestMapping` to `@PostMapping` but it's going to be easier to scan read if this is done consistently. Please either change them for all controllers in snowstorm using a separate refactoring commit or just use the current standard.

src/main/java/org/snomed/snowstorm/validation/ConceptValidationThenOperationService.java
```diff
@@ -0,0 +1,223 @@
...
+			return operation == ConceptValidationOperation.CREATE ? createConcept(concept, acceptLanguageHeader, branchPath, invalidContents.stream().filter(invalidContent ->
+					invalidContent.getSeverity() == Severity.WARNING).collect(Collectors.toList())) : updateConcept(concept, acceptLanguageHeader, branchPath,
+					invalidContents.stream().filter(invalidContent -> invalidContent.getSeverity() == Severity.WARNING).collect(Collectors.toList()));
+		}
+		throw new DroolsValidationException(objectMapper.writeValueAsString(invalidContentErrors).replace("\n", ""));
+	}
+
+	private void throwExceptionIfFieldsNotSet() {
+		if (concept == null || acceptLanguageHeader == null || branchPath == null || operation == null) {
+			throw new IllegalArgumentException("The mandatory field '" + (concept == null ? "concept" : (acceptLanguageHeader == null ? "acceptLanguageHeader"
```
kaicode, line 110: `acceptLanguageHeader` is not mandatory. I would use the `ControllerHelper.parseAcceptLanguageHeaderWithDefaultFallback` in the ConceptController before passing the result on to this service class, it will provide a default.

src/main/java/org/snomed/snowstorm/validation/ConceptValidationThenOperationService.java
```diff
@@ -0,0 +1,223 @@
...
+		}
+		return concept;
+	}
+
+	/**
+	 * Generates a random UUID with length == 10 characters.
+	 *
+	 * @return random UUID with character length of 10.
+	 */
+	public static String generateRandomUUID() {
```
kaicode, line 86: Please use `UUID.randomUUID().toString()` instead.

src/main/java/org/snomed/snowstorm/validation/ConceptValidationThenOperationService.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.Random;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+import static org.snomed.snowstorm.rest.ControllerHelper.getCreatedLocationHeaders;
+
+@Service
+public class ConceptValidationThenOperationService {
```
kaicode, line 38: I feel like this should be a helper class rather than a service class as it deals with some controller concerns rather than just business logic.
If we have ConceptValidationHelper with a method like `validate` that could run the validation and throw an exception if there are errors or return a validation header if there are warnings.
That way the final call to the ConceptService to create or update and the creation of the `ResponseEntity` could be in the `ConceptController` which would make for easy reading and keep controller concerns in one place as far as possible. 
This is a suggestion, what are your thoughts?

src/main/java/org/snomed/snowstorm/validation/ConceptValidationThenOperationService.java
```diff
@@ -0,0 +1,223 @@
...
+	public final ResponseEntity<ConceptView> execute() throws JsonProcessingException, ServiceException {
+		throwExceptionIfFieldsNotSet();
+		final List<InvalidContent> invalidContents = validationService.validateConcept(branchPath, concept);
+		final List<InvalidContent> invalidContentErrors = invalidContents.stream().filter(invalidContent -> invalidContent.getSeverity() == Severity.ERROR).collect(Collectors.toList());
+		if (invalidContentErrors.isEmpty()) {
+			return operation == ConceptValidationOperation.CREATE ? createConcept(concept, acceptLanguageHeader, branchPath, invalidContents.stream().filter(invalidContent ->
+					invalidContent.getSeverity() == Severity.WARNING).collect(Collectors.toList())) : updateConcept(concept, acceptLanguageHeader, branchPath,
+					invalidContents.stream().filter(invalidContent -> invalidContent.getSeverity() == Severity.WARNING).collect(Collectors.toList()));
+		}
+		throw new DroolsValidationException(objectMapper.writeValueAsString(invalidContentErrors).replace("\n", ""));
```
kaicode, line 105: Nice catch replacing the newlines. If this replacement was made while setting the header it would be clearer why it's happening, otherwise a comment would be good.

kaicode (commented):
I've suggested some changes and asked a question. Looking good so far!

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -11,9 +12,24 @@
...
 import org.snomed.snowstorm.config.Config;
-import org.snomed.snowstorm.core.data.domain.*;
+import org.snomed.snowstorm.core.data.domain.CodeSystem;
+import org.snomed.snowstorm.core.data.domain.CodeSystemVersion;
+import org.snomed.snowstorm.core.data.domain.Concept;
+import org.snomed.snowstorm.core.data.domain.ConceptMini;
+import org.snomed.snowstorm.core.data.domain.ConceptView;
+import org.snomed.snowstorm.core.data.domain.Description;
+import org.snomed.snowstorm.core.data.domain.QueryConcept;
+import org.snomed.snowstorm.core.data.domain.Relationship;
```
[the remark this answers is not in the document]
Tyler-Cooksey: Fixed.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -369,30 +413,45 @@ public ConceptReferencesResult findConceptReferences(
 		return new ConceptReferencesResult(typeSets, conceptReferencesPage.getPageable(), conceptReferencesPage.getTotalElements());
 	}
 
-	@RequestMapping(value = "/browser/{branch}/concepts", method = RequestMethod.POST)
+	@PostMapping(value = "/browser/{branch}/concepts/{validate}")
 	@PreAuthorize("hasPermission('AUTHOR', #branch)")
 	@JsonView(value = View.Component.class)
 	public ResponseEntity<ConceptView> createConcept(
 			@PathVariable String branch,
+			@PathVariable(required = false) Boolean validate,
```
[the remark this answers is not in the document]
Tyler-Cooksey: Fixed

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
Tyler-Cooksey: Fixed

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
Tyler-Cooksey: Fixed by refactoring the whole Controller to use the new annotation introduced in version 4.3 of Spring.

src/main/java/org/snomed/snowstorm/validation/ConceptValidationThenOperationService.java
Tyler-Cooksey: Fixed

src/main/java/org/snomed/snowstorm/validation/ConceptValidationThenOperationService.java
Tyler-Cooksey: Yes I agree this would make it more readable and will make the code changes in a separate commit (while I'll do now). Thanks.

Tyler-Cooksey: Fixed.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
kaicode: I suggested all snowstorm controllers but this is fine.

kaicode (approved):
Looks great now!

PR #218, opened 2021-02-09
Fix Kubernetes config

nightscape:
`java -jar snowstorm.jar` is already part of the `ENTRYPOINT` of the container.


kaicode (approved):
Thanks for pointing this out. I have no experience with kubernetes and this configuration does not get tested by the internal team but I understand that the kubernetes setup is pulling the snowstorm Docker image which already has these arguments in the ENTRYPOINT [here](https://github.com/IHTSDO/snowstorm/blob/5.1.0/Dockerfile#L28).

PR #219, opened 2021-02-09
Expose the used port 8080

nightscape:


kaicode (approved):
Good idea, this prevents the need to expose the default port manually.

PR #222, opened 2021-02-10
CDI-43 - Concrete values can now be stored within the stated form (axioms) when creating and updating concepts.

Tyler-Cooksey:
The underlying issue was related to marshalling, as Jackson requires a default no arg constructor and the ConcreteValue did not have one so it was throwing a mapping exception. After adding this default public constructor, the marshalling is performed successfully and the create/update operation work as expected.


kaicode (approved):
Looks good to me. 
Next time please use any concept id other than `116680003 | Is a (attribute) |` for the concrete relationship type as this concept has special meaning and special handling within the semantic index processing during a commit.

PR #223, opened 2021-02-11
Run build inside docker

nightscape:
This PR allows building Snowstorm without having a local Maven setup.
This is a prerequisite to automate Docker deployments using e.g. Github Actions.


kaicode:
@rorydavidson would this work for us? Can you think of any down sides?

rorydavidson:
> @rorydavidson would this work for us? Can you think of any down sides?

@kaicode This approach has not been a use case for us at the moment, but I can't see any harm in this on first look (obviously need to try it first). Would just mean tweaking out build processes for docker containers.

kaicode:
@rorydavidson I believe you have implemented this approach now so this PR could be closed?

rorydavidson:
This was incorporated and the docker file now includes the build process per this pull request.

PR #226, opened 2021-02-16
MAINT-1462 - Refactored the code to insert the validation results into the body of the Concept when ERRORS/WARNINGS are found after validation has been performed and moved the Spring validation around so that it performs the validation after the creation/update operation has been performed when the validate flag has been set to true. If it's set to false, it will simply perform the validation first (as normal) then proceed with the underlying operation.

Tyler-Cooksey:
Example Response of the validation results as part of the Concept object returned which contains ERRORS/WARNINGS:
image

### DIFF
--- a/src/main/java/org/snomed/snowstorm/rest/ConceptController.java
+++ b/src/main/java/org/snomed/snowstorm/rest/ConceptController.java
@@ -2,7 +2,6 @@ package org.snomed.snowstorm.rest;
 
 import ch.qos.logback.classic.Level;
 import com.fasterxml.jackson.annotation.JsonView;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.kaicode.elasticvc.api.BranchCriteria;
 import io.kaicode.elasticvc.api.VersionControlHelper;
@@ -12,7 +11,7 @@ import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
 import org.elasticsearch.common.Strings;
-import org.ihtsdo.drools.response.InvalidContent;
+import org.ihtsdo.drools.response.Severity;
 import org.snomed.snowstorm.config.Config;
 import org.snomed.snowstorm.core.data.domain.*;
 import org.snomed.snowstorm.core.data.domain.expression.Expression;
@@ -29,6 +28,7 @@ import org.snomed.snowstorm.ecl.validation.ECLValidator;
 import org.snomed.snowstorm.rest.converter.SearchAfterHelper;
 import org.snomed.snowstorm.rest.pojo.*;
 import org.snomed.snowstorm.validation.ConceptValidationHelper;
+import org.snomed.snowstorm.validation.ConceptValidationHelper.InvalidContentWithSeverityStatus;
 import org.snomed.snowstorm.validation.DroolsValidationService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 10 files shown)


src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -387,17 +391,24 @@ public ConceptReferencesResult findConceptReferences(
...
-			final Concept createdConcept = conceptService.create((Concept) concept, ControllerHelper.parseAcceptLanguageHeaderWithDefaultFallback(acceptLanguageHeader),
-					BranchPathUriUtil.decodePath(branch));
-			return new ResponseEntity<>(createdConcept, ConceptValidationHelper.getCreateConceptHeaders(ControllerHelper.getCreatedLocationHeaders(createdConcept.getId()),
-					objectMapper.writeValueAsString(ConceptValidationHelper.replaceTemporaryUUIDWithSCTID(invalidContentWarnings, createdConcept))), HttpStatus.OK);
+			final InvalidContentWithSeverityStatus invalidContent = ConceptValidationHelper.validate((Concept) concept, BranchPathUriUtil.decodePath(branch), validationService);
+			if (invalidContent.getSeverity() == Severity.WARNING) {
+				final Concept createdConcept = conceptService.create((Concept) concept, ControllerHelper.parseAcceptLanguageHeaderWithDefaultFallback(acceptLanguageHeader),
+						BranchPathUriUtil.decodePath(branch));
+				// validate concept after creation due to temporary id provided by UI can exceed the concept id range.
+				validator.validate(createdConcept);
```
kaicode, line 403: Is this just the simple bean validation, not the expensive drools call? It's not clear when reading the code, maybe the naming could be adjusted somehow?

src/main/java/org/snomed/snowstorm/validation/ConceptValidationHelper.java
```diff
@@ -23,42 +18,29 @@
...
-	public static List<InvalidContent> validate(final Concept concept, final String branchPath, final DroolsValidationService validationService,
-			final ObjectMapper objectMapper) throws JsonProcessingException, ServiceException {
+	public static InvalidContentWithSeverityStatus validate(final Concept concept, final String branchPath, final DroolsValidationService validationService) throws ServiceException {
 		final List<InvalidContent> invalidContents = validationService.validateConcept(branchPath, generateUUIDSIfNotSet(concept));
 		final List<InvalidContent> invalidContentErrors = invalidContents.stream().filter(invalidContent -> invalidContent.getSeverity() == Severity.ERROR).collect(Collectors.toList());
-		if (invalidContentErrors.isEmpty()) {
-			return invalidContents.stream().filter(invalidContent -> invalidContent.getSeverity() == Severity.WARNING).collect(Collectors.toList());
-		}
-		throw new DroolsValidationException(objectMapper.writeValueAsString(invalidContentErrors).replace("\n", ""));
+		return invalidContentErrors.isEmpty() ? new InvalidContentWithSeverityStatus(invalidContents.stream().filter(invalidContent ->
```
kaicode, line 24: The filter probably isn't required here? Because if there are no errors we can assume the list only contains warnings.

kaicode (changes requested):
Looking good. Made a couple of suggestions, should be quick to fix.

src/main/java/org/snomed/snowstorm/validation/ConceptValidationHelper.java
Tyler-Cooksey: Nice spot, fixed.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
Tyler-Cooksey: Good suggestion, fixed.

kaicode (approved):
Looks good to me 👍

PR #230, opened 2021-02-22
MAINT-1323 Fix concepts losing release details

pgwilliams:
Scenario for MAINT-1323 is that a concept is versioned on MAIN and also edited on some task.   The versioned concept is rebased to the project and the task is promoted, causing a merge conflict.    The user selects the right hand side and, during the merge, the "existing" concept that is used to copy the "release details" is taken from the task branch.  The details relating to versioning are then lost.     This code fix additionally collects the existing concepts from the _source_ branch (in this case the project) and pulls the release details from there **if** a concept has been released when the task version of the concept has not been.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
@@ -608,7 +608,15 @@ public class ConceptService extends ComponentService {
 			return new PersistedComponents();
 		}
 
-		PersistedComponents persistedComponents = conceptUpdateHelper.saveNewOrUpdatedConcepts(concepts, commit, getExistingConceptsForSave(concepts, commit));
+		Map<String, Concept> existingTargetConceptsMap = getExistingConceptsForSave(concepts, commit);
+		
+		//Populate source concepts on rebase
+		Map<String, Concept> existingSourceConceptsMap = null;
+		if (commit.getCommitType().equals(Commit.CommitType.REBASE)) {
+			existingSourceConceptsMap = getExistingSourceConceptsForSave(concepts, commit);
+		}
+		
+		PersistedComponents persistedComponents = conceptUpdateHelper.saveNewOrUpdatedConcepts(concepts, commit, existingTargetConceptsMap, existingSourceConceptsMap);
 
 		// Log traceability activity
 		if (logTraceability && traceabilityLogService.isEnabled()) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -734,6 +742,21 @@ private void joinComponentsToConcepts(PersistedComponents persistedComponents, M
 		}
 		return existingConceptsMap;
 	}
+	
+	private Map<String, Concept> getExistingSourceConceptsForSave(Collection<Concept> concepts, Commit commit) {
+		Map<String, Concept> existingSourceConceptsMap = new HashMap<>();
+		final List<String> conceptIds = concepts.stream().filter(concept -> concept.getConceptId() != null).map(Concept::getConceptId).collect(Collectors.toList());
+		if (!conceptIds.isEmpty()) {
+			BranchTimepoint branchTimePoint = new BranchTimepoint(commit.getSourceBranchPath(), BranchTimepoint.DATE_FORMAT.format(commit.getTimepoint()));
```
kaicode, line 750: It would be safer to use `commit.getBranch().getBase()` rather than timepoint because it's possible to rebase onto a point on the parent branch which is not the latest, for example when upgrading an extension.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -81,7 +81,9 @@
 
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
-	PersistedComponents saveNewOrUpdatedConcepts(Collection<Concept> concepts, Commit commit, Map<String, Concept> existingConceptsMap) throws ServiceException {
+	PersistedComponents saveNewOrUpdatedConcepts(Collection<Concept> concepts, Commit commit, 
+			Map<String, Concept> existingTargetConceptsMap,
```
kaicode, line 85: Could we name this back to `existingConceptsMap` because that's more intuitive and they are the existing concepts on this branch.
The other map could be called `rebaseExistingSourceConceptsMap` too to make it more obvious?

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -133,10 +137,17 @@ PersistedComponents saveNewOrUpdatedConcepts(Collection<Concept> concepts, Commi
 			// Mark changed concepts as changed
 			if (existingConcept != null) {
 				concept.setCreating(false);// May have been set true earlier during first save
+				
+				//Are we copying the released details from the existing concept, or has the source been versioned recently?
+				Concept authoratativeConcept = existingConcept;
```
kaicode, line 142: It would be cleaner to do this within a new version of the `SnomedComponent.copyReleaseDetails` method that takes both the existingConcept and existingSourceConcept. The same should apply for the other component types too!
Also I think this logic should take the component with the greatest 'getReleasedEffectiveTime()'.

kaicode (changes requested):
The unit test plugs a real gap in coverage 👍 
I've made a couple of minor suggestions for readability and covering other types of components (descriptions, relationships and axioms).

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
pgwilliams: I think I tried that and ended up receiving the copy of the concept before it had been versioned and it had the end element set on it.   But I'm not sure which branch base I used there, so I'll give that a test and see if it still passes.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
pgwilliams: Agree.   I've tried to keep that same existingRebaseSourceComponent naming througout.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
pgwilliams: OK, yes of course I should have done this for all component types.    I've created public void copyReleaseDetails(SnomedComponent<C> component, SnomedComponent<C> existingRebaseSourceComponent)  and then I commented out the previous one temporarily to see all the places it was being called and switched over to the rebase version where possible. 

The only place I wasn't sure about (because it gets called from A LOT) of places is ReferenceSetMemberService @ 374 which I suspect needs to have an optional rebaseBranch passed in so we can call our new function at line 380 if we have more than one existing member to consider.

kaicode (approved):
This is brilliant! Thank you @pgwilliams

PR #234, opened 2021-02-26
MAINT-1462 - Removed validator as the validation operation is performed already, as part of the create/update operation. I also added in a method to strip out the temporary UUIDs after the validation has been performed and the result only contains warnings so then it can perform the underlying create/update operation successfully.

Tyler-Cooksey:


kaicode (approved):
Looks good. Let's cast concept to Concept once on the first line of the controller then we don't need to keep casting it every time we need it. Feel free to merge after that 👍

PR #235, opened 2021-02-26
Maint 1346

CoderMChu:
@kaicode  Please review my changes for this. I have updated the API parameters after our discussion. However I have decided to leave the metadata updating logic out for now when creating a new code system. Because Dev-ops still need to run the metadata update for other configs.


kaicode (approved):
Looks good to me 👍

PR #237, opened 2021-03-03
MAINT-1588 Updated duplicate components fix logic during extension up…

CoderMChu:
…grade to prevent extension components being overwritten


kaicode (approved):
Looks good to me 👍

PR #238, opened 2021-03-08
CDI-107 - Better handling of ECL queries for edge cases

Tyler-Cooksey:


src/main/java/org/snomed/snowstorm/ecl/ECLQueryService.java
```diff
@@ -39,8 +43,15 @@
 	}
 
 	public Page<Long> selectConceptIds(String ecl, BranchCriteria branchCriteria, String path, boolean stated, Collection<Long> conceptIdFilter, PageRequest pageRequest) throws ECLException {
+		return selectRelevantConceptIds(ecl, branchCriteria, path, stated, conceptIdFilter, pageRequest, eclEdgeCaseHandlerService.replaceIncorrectConcreteAttributeValue(ecl, path, pageRequest));
```
kaicode, line 46: Consider putting the `replaceIncorrectConcreteAttributeValue` call on a separate line for readability. This is a significant step in processing. This style would be fine for some sort of utility method call.

src/main/java/org/snomed/snowstorm/ecl/ECLQueryService.java
```diff
@@ -39,8 +43,15 @@
...
+	}
+
+	public Page<Long> selectRelevantConceptIds(String ecl, BranchCriteria branchCriteria, String path, boolean stated, Collection<Long> conceptIdFilter, PageRequest pageRequest, SExpressionConstraint expressionConstraint) {
```
kaicode, line 49: This is really the same as `selectConceptIds` but with different arguments. It's more like an internal call because the ECL has already been parsed. Consider naming this `doSelectConceptIds` which is common naming for a lower level call after parsing / processing of the arguments.
I think this method should require the expressionConstraint parameter, otherwise this method could be used to skip the `replaceIncorrectConcreteAttributeValue` behaviour without meaning to. If that is the intention maybe this intension could be reflected in the naming of the method instead?

src/main/java/org/snomed/snowstorm/ecl/validation/ECLEdgeCaseHandlerService.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static org.snomed.snowstorm.core.data.domain.Concepts.CONCEPT_MODEL_DATA_ATTRIBUTE;
+import static org.snomed.snowstorm.ecl.domain.expressionconstraint.SExpressionConstraintHelper.MISSING;
+
+@Service
+public class ECLEdgeCaseHandlerService implements CommitListener {
```
kaicode, line 28: Running every query through a service named "edge case" seems counter intuitive. Maybe naming like `ECLPreprocessingService` would show the intent more clearly?

src/main/java/org/snomed/snowstorm/ecl/validation/ECLEdgeCaseHandlerService.java
```diff
@@ -0,0 +1,131 @@
...
+
+	private static final String RETURN_ALL_CONCRETE_ATTRIBUTES_ECL_QUERY = "< " + CONCEPT_MODEL_DATA_ATTRIBUTE;
+
+	private static final List<String> CACHED_CONCRETE_CONCEPT_IDS = new ArrayList<>();
+
+	@Autowired
+	private ECLQueryBuilder eclQueryBuilder;
```
kaicode, line 35: It would be better separation of concerns if this class took SExpressionConstraint as input and output. It does not need to be involved in calling the ECL builder.

src/main/java/org/snomed/snowstorm/ecl/validation/ECLEdgeCaseHandlerService.java
```diff
@@ -0,0 +1,131 @@
...
+
+	@Override
+	public final void preCommitCompletion(final Commit commit) throws IllegalStateException {
+		CACHED_CONCRETE_CONCEPT_IDS.clear();
+	}
+
+	public SExpressionConstraint replaceIncorrectConcreteAttributeValue(final String ecl, final String branch, final PageRequest pageRequest) {
+		SExpressionConstraint expressionConstraint = null;
+		if (ecl != null) {
+			if (CACHED_CONCRETE_CONCEPT_IDS.isEmpty()) {
```
kaicode, line 51: This cache must be branch specific.

src/main/java/org/snomed/snowstorm/ecl/validation/ECLEdgeCaseHandlerService.java
```diff
@@ -0,0 +1,131 @@
...
+
+	private void setPlaceholder(final SEclAttribute sEclAttribute) {
+		if (sEclAttribute.getNumericValue() != null) {
+			sEclAttribute.setNumericValue(String.valueOf(Float.MAX_VALUE));
+		} else if (sEclAttribute.getStringValue() != null) {
+			sEclAttribute.setStringValue(MISSING);
+		}
+	}
+
+	private List<SEclAttribute> extractSEclAttributes(final SRefinedExpressionConstraint expressionConstraint) {
```
kaicode, line 72: We will need to call a method like this recursively to cover all possible cases. The logic here may cover the unit tests but ECL can contain nested queries to any level. I can help with this if needed.

src/main/java/org/snomed/snowstorm/ecl/validation/ECLEdgeCaseHandlerService.java
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	private void cacheConcreteConceptIds(final String branch, final PageRequest pageRequest) {
+		CACHED_CONCRETE_CONCEPT_IDS.addAll(eclQueryService.selectRelevantConceptIds(RETURN_ALL_CONCRETE_ATTRIBUTES_ECL_QUERY,
+				versionControlHelper.getBranchCriteria(branch), branch, false, null, pageRequest, null)
+				.toList().stream().map(Functions.toStringFunction()).collect(Collectors.toList()));
+	}
+
+	@Override
+	public boolean equals(Object o) {
```
kaicode, line 118: `@Service` classes should not need to implement `equals` and `hashCode`. These are only needed when the object is put into a collection or compared with another instance of the same class. There should only ever be one instance of a `@Service` class.

src/test/java/org/snomed/snowstorm/core/data/services/SemanticIndexUpdateServiceTest.java
```diff
@@ -1066,71 +1066,69 @@ void testRebuildSemanticIndexWithConcreteValues() throws ServiceException {
...
+		assertEquals(0, queryService.search(queryService.createQueryBuilder(false).ecl("*:396070081 = \"50\""), path, QueryService.PAGE_OF_ONE_HUNDRED).getTotalElements());
+		assertEquals(0, queryService.search(queryService.createQueryBuilder(false).ecl("*:396070081 = \"#50\""), path, QueryService.PAGE_OF_ONE_HUNDRED).getTotalElements());
 
 		// number query
-		assertEquals(1, queryService.search(queryService.createQueryBuilder(false).ecl("*:396070080 = #100.000005"), path, QueryService.PAGE_OF_ONE).getTotalElements());
-		assertEquals(1, queryService.search(queryService.createQueryBuilder(false).ecl("*:396070080 = #50"), path, QueryService.PAGE_OF_ONE).getTotalElements());
-		assertEquals(0, queryService.search(queryService.createQueryBuilder(false).ecl("*:396070080 = #20"), path, QueryService.PAGE_OF_ONE).getTotalElements());
-		assertEquals(1, queryService.search(queryService.createQueryBuilder(false).ecl("*:396070082 = #20"), path, QueryService.PAGE_OF_ONE).getTotalElements());
+		assertEquals(1, queryService.search(queryService.createQueryBuilder(false).ecl("*:396070080 = #100.000005"), path, QueryService.PAGE_OF_ONE_HUNDRED).getTotalElements());
```
kaicode, line 1097: A page of 1 is sufficient if we are just reading the total elements from the page.

kaicode (changes requested):

kaicode:
@Tyler-Cooksey A good start on a tricky feature. I've made a couple of comments for your consideration. Cheers.

kaicode (approved):
Looks good to me now 👍

CoderMChu (approved):
@Tyler-Cooksey Good work.

PR #239, opened 2021-03-08
MAINT-1372 Fix Snowstorm Log error "No value specified for terms query"

dmcgihtsdo:
MAINT-1372 is concerned with fixing a bug in Snowstorm where the "No value specified for terms query" message appears in the log. 

To reproduce this bug, send a POST request to the _concepts/bulk-load_ endpoint with null Concept identifiers; the request will fail. The request will fail because the queryBuilder (ConceptService, line 359) will attempt to build a query with an invalid (null) terms query. 

To fix this bug, I am ignoring all null identifiers (by deserialising with setConceptIdsSafely).


kaicode (approved):
This is awesome: `conceptIds.removeIf(Objects::isNull)`.
Should we do the same for `descriptionIds`?

kaicode (approved):

PR #242, opened 2021-03-12
MAINT-386 Fixed issue that merge allows duplicate PTs

QuyenLy87:


kaicode (changes requested):
This looks like it will work but... although it's a nice idea to include the acceptability additional field in the first fetch of the refset members, this is likely to have a performance impact for large rebases. The reason for this is that the ReferenceSetMember `referencedComponentId` and `conceptId` fields are marked as stored fields in elasticsearch which makes them extremely fast to fetch from the store. Fetching the `additionalFields` for a large number of reference sets will be very expensive so if we can avoid that we should.
Could you just fetch the concept id and reference component id in the first fetch then make a second fetch of just the reference component id field with a query requiring the referenced component to be in the synonymAndTextDefIds and the acceptability of preferred. Checking the acceptability within Elasticsearch is really fast, then we just ask for the reference component id back.

This is just something I have learnt through experience and we are trying to keep Snowstorm as fast as possible. Thank you for your patience.
The unit test looks great!

QuyenLy87:
Hi @kaicode , Looks like we need one more query to find the preferred description ids:
**NativeSearchQuery languageMemberQuery = newSearchQuery(updatesDuringRange)
				.withFilter(boolQuery()
						.must(existsQuery(ReferenceSetMember.Fields.CONCEPT_ID))
						.must(termsQuery(ReferenceSetMember.Fields.REFERENCED_COMPONENT_ID, synonymAndTextDefIds))
						.must(termsQuery(ReferenceSetMember.LanguageFields.ACCEPTABILITY_ID_FIELD_PATH, Concepts.PREFERRED)))
				.withFields(ReferenceSetMember.Fields.REFERENCED_COMPONENT_ID)
				.build();
		try (final SearchHitsIterator<ReferenceSetMember> stream = elasticsearchTemplate.searchForStream(languageMemberQuery, ReferenceSetMember.class)) {
			stream.forEachRemaining(hit -> preferredDescriptionIds.add(parseLong(hit.getContent().getReferencedComponentId())));
		}**

Is that enough?

kaicode:
@QuyenLy87 yes, that looks good.

QuyenLy87:
Hi @kaicode please review this pull request again. Thanks

kaicode (approved):
Perfect!

PR #246, opened 2021-03-19
MAINT-1572 Reduce Concept save time

dmcgihtsdo:
MAINT-1572 is concerned with reducing Concept save time by prefetching SCTIDs for non-International namespaces.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
@@ -6,6 +6,7 @@ import org.snomed.snowstorm.core.data.domain.ComponentType;
 import org.snomed.snowstorm.core.data.services.RuntimeServiceException;
 import org.snomed.snowstorm.core.data.services.ServiceException;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
@@ -31,6 +32,9 @@ public class IdentifierCacheManager implements Runnable {
 
 	@Autowired
 	private IdentifierSource identifierSource;
+
+	@Value("${cis.cache.concept-prefetch-count}")
+	private int conceptIdPrefetchCount;
 
 	// Separate cache for each namespace/partition combination configured.
 	private Set<IdentifierCache> identifierCaches = new HashSet<>();
@@ -154,6 +158,24 @@ public class IdentifierCacheManager implements Runnable {
 				cache.unlock();
 				requestSatisfied = true;
 			}
+		} else {
+			//If no cache available & not requesting for International (as already prefetched),
+			//then prefetch additional identifiers for subsequent requests.
+			int prefetchQuantity;
+			switch (componentType) {
+				case Concept:
+					prefetchQuantity = conceptIdPrefetchCount;
+					break;
+				case Description:
+					prefetchQuantity = conceptIdPrefetchCount * 2;
+					break;
+				case Relationship:
+					prefetchQuantity = conceptIdPrefetchCount * 4;
+					break;
+				default:
+					throw new IllegalArgumentException("Cache does not support prefetching identifiers for " + componentType);
+			}
+			addCache(namespaceId, partitionId, prefetchQuantity);
 		}
 		
 		if (!requestSatisfied) {
@@ -194,6 +216,7 @@ public class IdentifierCacheManager implements Runnable {
 		stayAlive = false;
 		cacheDaemon.interrupt();
 		cacheDaemon = null;
+		identifierCaches.clear();
 	}
 
 
--- a/src/test/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManagerTest.java
+++ b/src/test/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManagerTest.java
@@ -2,6 +2,7 @@ package org.snomed.snowstorm.core.data.services.identifier;
 
 import org.junit.Assert;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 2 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
```diff
@@ -154,6 +158,24 @@ public void populateIdBlock(IdentifierReservedBlock idBlock, int quantityRequire
 				cache.unlock();
 				requestSatisfied = true;
 			}
+		} else {
+			//If no cache available & not requesting for International (as already prefetched),
+			//then prefetch additional identifiers for subsequent requests.
+			int prefetchQuantity;
+			switch (componentType) {
+				case Concept:
+					prefetchQuantity = conceptIdPrefetchCount;
```
dmcgihtsdo, line 167: @kaicode , I have copied the prefetch quantity for Concepts, Descriptions and Relationships from Config::getIdentifierCacheManager. Are you happy with the numbers?

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
```diff
@@ -194,6 +216,7 @@ public void stopBackgroundTask() {
 		stayAlive = false;
 		cacheDaemon.interrupt();
 		cacheDaemon = null;
+		identifierCaches.clear();
```
dmcgihtsdo, line 219: I added this line because the tests in IdentifierCacheManagerTest were interfering with each other.

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
```diff
@@ -154,6 +158,24 @@ public void populateIdBlock(IdentifierReservedBlock idBlock, int quantityRequire
 				cache.unlock();
 				requestSatisfied = true;
 			}
+		} else {
+			//If no cache available & not requesting for International (as already prefetched),
+			//then prefetch additional identifiers for subsequent requests.
+			int prefetchQuantity;
```
kaicode, line 164: When we create a cache it would be good to create for all component types in the namespace, as we do for International on startup

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
kaicode, line 219: Do we need this? Doesn't that kill the caches?

src/test/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManagerTest.java
```diff
@@ -65,6 +66,86 @@ void testTopUp() throws ServiceException, InterruptedException {
 		Long sctid = reservedBlock.getNextId(ComponentType.Concept);
 		Assert.assertNull(IdentifierService.isValidId(sctid.toString(), ComponentType.Concept));
 	}
+
+	@Test
+	void populateIdBlock_ShouldReadFromCache_WhenRequestingConceptIdentifierWithNonInternationalNamespace() throws ServiceException {
```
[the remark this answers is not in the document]
dmcgihtsdo: Yes they do. They will fail from a NPE as the cache doesn't have an entry for the given namespace.

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
dmcgihtsdo: It does kill the cache. I made the change as the cache in the tests were overlapping. Because the method is only called from the shutdownPolling method, I figured it was fine to do as that method itself is called when the bean is no longer required.

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
dmcgihtsdo: @kaicode , I can update my code to include Axioms, but I'm not sure where you're seeing Axioms being included for International. In Config::getIdentifierCacheManager, only Concepts, Descriptions and Relationships are added to the cache. Shall I update that to include Axioms?

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
kaicode: yeah, the numbers are fine thanks

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
kaicode: I didn't realised this was within `stopBackgroundTask`, that's fine 👍

src/main/java/org/snomed/snowstorm/core/data/services/identifier/IdentifierCacheManager.java
kaicode: There are no SCTIDs for axioms. Maybe I am not reading this correctly but it looks like this block only creates a cache for the component type being requested. I think if any type is requested we could create a cache and prefetch ids for all types straight away.

dmcgihtsdo: @kaicode , ah I get you now. Great spot. Yes, you are correct in that the cache is only updated for one given component type. However, the `idBlock` variable is from `IdentifierService::getReservedBlock` which calls the method for all of the component types: 

```
cacheManager.populateIdBlock(idBlock, conceptIds, namespace, ...);
cacheManager.populateIdBlock(idBlock, descriptionIds, namespace, ...);
cacheManager.populateIdBlock(idBlock, relationshipIds, namespace, ...);
```

kaicode: Got it. This looks good then 😄

kaicode (approved):

PR #247, opened 2021-03-23
MAINT-1624 Sort Relationships before responding with Concept preview

dmcgihtsdo:
MAINT-1624 is concerned with updating the Classification's before/after view such that it is the same as the editing view panel.


kaicode (approved):
Looks good to me 👍

PR #255, opened 2021-04-13
Fix AWS signing with empty request entity

rowanvb:
amazon-archives/aws-request-signing-apache-interceptor#8

This change hasn't been published to the maven repository, which seems to not be maintained by anyone anymore. https://mvnrepository.com/artifact/com.github.awslabs/aws-request-signing-apache-interceptor/deb7941e85
amazon-archives/aws-request-signing-apache-interceptor#2

Props to Amazon for not including this in their java sdk 😅

I am not entirely sure if it's preferable to have this interceptor in this repository, but the alternatives are to try to find someone who maintains this dependency in a public repository, with no guarantees that he will keep maintaining this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,6 +36,7 @@
 
 		<docker.image.prefix>snomedinternational</docker.image.prefix>
 		<slf4j.version>1.7.5</slf4j.version>
+		<aws-java-sdk.version>1.11.221</aws-java-sdk.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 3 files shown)


rowanvb:
During startup, some (POST) requests already fail. I'll look them up tomorrow. Also HEAD requests on /description fail, so it seems not to be limited to just empty body POST requests. Without this empty content, the signature is invalid, so AWS denies the requests. Haven't been able to pinpoint where exactly this problem started, since we've been stuck on version 4.4.0 for a while.

kaicode:
Thank you for this PR. I've scheduled some time to take a look at this tomorrow.

pom.xml
```diff
@@ -36,6 +36,7 @@
 
 		<docker.image.prefix>snomedinternational</docker.image.prefix>
 		<slf4j.version>1.7.5</slf4j.version>
+		<aws-java-sdk.version>1.11.221</aws-java-sdk.version>
```
kaicode, line 39: This explicit version dependency matches the version that was pulled in via `aws-request-signing-apache-interceptor` previously 👍

kaicode (approved):
I think brining in this single class (less than 200 lines) with the bug fix rather than using the dependency which is no longer maintained is a good approach.
Thank you for providing this fix.

kaicode:
We will include this in the next Snowstorm release. It will be a new major version (7.0.x) using Java 11.
Many thanks!

PR #256, opened 2021-04-14
MAINT-1030 FHIR: Set inactive flag by default when concept is inactive

dmcgihtsdo:
MAINT-1030 is concerned with updating the response for both CodeSystem $lookup and ValueSet $expand to include the Concept's active state.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/fhir/services/FHIRHelper.java
+++ b/src/main/java/org/snomed/snowstorm/fhir/services/FHIRHelper.java
@@ -1,19 +1,23 @@
 package org.snomed.snowstorm.fhir.services;
 
-import java.text.ParseException;
-import java.util.*;
-
-import javax.servlet.http.HttpServletRequest;
-
+import ca.uhn.fhir.rest.param.StringParam;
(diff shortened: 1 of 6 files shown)


pgwilliams (approved):
Super.   Also appreciate the unit testing there.

pgwilliams (commented):
I'm not clear about the decisions around when to use a .* for your imports and when to list them out individually.    Is your IDE handling that for you, or do you have some convention you follow?   Not a big deal, just interested.

Also checking to see if I can find a unit test relating to the combination of ECL + explicit concept inclusion.   Will look at your other commit...

dmcgihtsdo:
> I'm not clear about the decisions around when to use a .* for your imports and when to list them out individually. Is your IDE handling that for you, or do you have some convention you follow? Not a big deal, just interested.
> 
> Also checking to see if I can find a unit test relating to the combination of ECL + explicit concept inclusion. Will look at your other commit...

@pgwilliams , the imports are handled by my IDE. Is there a particular style/rule I should be using / you prefer using?

PR #257, opened 2021-04-21
MAINT-1268 Display additional dependantVersionEffectiveTime field

dmcgihtsdo:
MAINT-1268 is concerned with updating the response from the `/codesystems` endpoint to include the field `dependantVersionEffectiveTime` in the `latestVersion` block.


src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
```diff
@@ -310,6 +310,11 @@ private void copyDetailsFromCacheEntry(CodeSystem codeSystem, Pair<Date, CodeSys
 		codeSystem.setLanguages(cachedCodeSystem.getLanguages());
 		codeSystem.setModules(cachedCodeSystem.getModules());
 		codeSystem.setDependantVersionEffectiveTime(cachedCodeSystem.getDependantVersionEffectiveTime());
+		CodeSystemVersion latestVersion = codeSystem.getLatestVersion();
+		CodeSystemVersion cachedLatestVersion = cachedCodeSystem.getLatestVersion();
+		if (latestVersion != null && cachedLatestVersion != null) {
+			latestVersion.setDependantVersionEffectiveTime(cachedLatestVersion.getDependantVersionEffectiveTime());
+		}
```
dmcgihtsdo, line 317: When requesting Code Systems, the data is read from the cache. This block simply maps the cache to the response.

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
```diff
@@ -381,6 +386,12 @@ private synchronized void doJoinContentInformation(CodeSystem codeSystem, String
 		}
 
 		// Add to cache
+		CodeSystemVersion previousVersion = findPreviousVisibleVersion(codeSystem.getShortName());
```
dmcgihtsdo, line 389: This block is called on start up and populates the cache.

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
```diff
@@ -484,6 +495,14 @@ public CodeSystemVersion findLatestVisibleVersion(String shortName) {
 		return null;
 	}
 
+	public CodeSystemVersion findPreviousVisibleVersion(String shortName) {
```
dmcgihtsdo, line 498: The List returned from `findAllVersions(...)` is ordered and so I felt confident returning the first index. This is a similar approach to the `findLatestVisibleVersion` method.

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
```diff
@@ -484,6 +495,14 @@ public CodeSystemVersion findLatestVisibleVersion(String shortName) {
...
+		List<CodeSystemVersion> versions = findAllVersions(shortName, false, latestVersionCanBeFuture);
+		if (versions != null && versions.size() > 1) {
```
dmcgihtsdo, line 500: I have set the condition to be greater than 1 as there were some Code Systems (LOINC for example) that only had 1. Therefore, if a Code System doesn't have more than 1, then return `null` as that essentially indicates there is no previous version.

src/test/java/org/snomed/snowstorm/rest/CodeSystemControllerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private String listCodeSystems(String branchPath) {
+        return "http://localhost:" + port + "/codesystems/?forBranch=" + branchPath;
+    }
+
+    private Integer getDependantVersionEffectiveTime(ItemsPage<ItemsPage<?>> result) {
+        Collection<ItemsPage<?>> items = result.getItems();
+
+        if (items instanceof ArrayList) {
```
dmcgihtsdo, line 94: If you look at this if-block in your IDE, you'll see a lot of warnings about generics and unchecked assignments. I couldn't work out the correct way to remove them and I could only shift the problem and not resolve it. If you wouldn't mind, it would be great seeing your solution.

kaicode: I've provided two solutions here.. one low level and one using domain objects..
On line 79..
```
        ItemsPage<?> result = this.restTemplate.getForObject(requestUrl, ItemsPage.class);
```
Then 
```
    private Integer getDependantVersionEffectiveTime(ItemsPage<?> result) {
    	@SuppressWarnings("unchecked")
		List<Map<String, Object>> items = (List<Map<String, Object>>) result.getItems();
		Map<String, Object> item = items.get(0);
		@SuppressWarnings("unchecked")
		Map<String, Object> latestVersion = (Map<String, Object>) item.get("latestVersion");
		return (Integer) latestVersion.get("dependantVersionEffectiveTime");
    }
```
Using `@SuppressWarnings` is a bit of a workaround for when you really don't know the java type that you're dealing with up front, or in this case where the Map contains values with a mixture of classes, and need to cast.

The other approach is to use a ParameterizedTypeReference to get the response mapped straight into the domain objects like this:
```
ResponseEntity<ItemsPage<CodeSystem>> result = this.restTemplate.exchange(requestUrl, HttpMethod.GET, null, new ParameterizedTypeReference<>() {});
```
I haven't tested either of these examples but you should get the idea.

kaicode (changes requested):
For the record - Darryn and I had a chat about the requirements. More changes are coming.

dmcgihtsdo:
I am closing this PR and have submitted (https://github.com/IHTSDO/snowstorm/pull/265) instead to address MAINT-1268.

PR #261, opened 2021-04-29
RT2-56: export refset-only RF2

shapirod2:
RT2-56: export refset-only RF2

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/rf2/export/ExportService.java
+++ b/src/main/java/org/snomed/snowstorm/core/rf2/export/ExportService.java
@@ -64,7 +64,8 @@ public class ExportService {
 	@Autowired
 	private CodeSystemService codeSystemService;
 
-	private final Set<String> refsetTypesRequiredForClassification = Sets.newHashSet(Concepts.REFSET_MRCM_ATTRIBUTE_DOMAIN, Concepts.OWL_EXPRESSION_TYPE_REFERENCE_SET);
+	private final Set<String> refsetTypesRequiredForClassification = Sets
+			.newHashSet(Concepts.REFSET_MRCM_ATTRIBUTE_DOMAIN, Concepts.OWL_EXPRESSION_TYPE_REFERENCE_SET);
 
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/snomed/snowstorm/core/rf2/export/ExportService.java
```diff
@@ -235,15 +252,102 @@ private File exportRF2ArchiveFile(String branchPath, String filenameEffectiveDat
 			branchService.unlock(branchPath);
 		}
 	}
+	
+	private File exportReferenceSetsOnlyRF2ArchiveFile(String branchPath, String filenameEffectiveDate,
```
kaicode, line 256: Is it necessary to duplicate such a large portion of code? Putting some logic in the existing export method to skip concepts, descriptions and relationships if it's a refset only export would be a lot cleaner.
Cloning large methods means more maintenance and potential bugs in the future.

src/main/java/org/snomed/snowstorm/core/rf2/export/ExportService.java
```diff
@@ -252,27 +356,31 @@ private BoolQueryBuilder getContentQuery(RF2Type exportType, Set<String> moduleI
...
+				+ format("%s_%s_%s.txt", exportType.getName(), codeSystemRF2Name, filenameEffectiveDate);
 		logger.info("Exporting file {}", componentFilePath);
 		try {
 			// Open zip entry
 			zipOutputStream.putNextEntry(new ZipEntry(componentFilePath));
 
 			// Stream components into zip
-			try (ExportWriter<T> writer = getExportWriter(componentClass, zipOutputStream, extraFieldNames, entryFilenamePrefix.contains("Concrete"));
-					SearchHitsIterator<T> componentStream = elasticsearchTemplate.searchForStream(getNativeSearchQuery(contentQuery), componentClass)) {
+			try (ExportWriter<T> writer = getExportWriter(componentClass, zipOutputStream, extraFieldNames,
```
kaicode, line 380: I don't think you meant to change any of this.. ?

src/test/java/org/snomed/snowstorm/core/rf2/export/ExportServiceTest.java
```diff
@@ -482,7 +486,46 @@ void exportRF2ArchiveForClassification() throws Exception {
...
+		
+		String descriptionTypeRefsetMemberId = referenceSetMemberService.findMembers("MAIN", Concepts.FSN, PAGE_OF_ONE).getContent().get(0).getMemberId();
+		String descriptionLanguageRefsetMemberId = referenceSetMemberService.findMembers("MAIN", descriptionId, PAGE_OF_ONE).getContent().get(0).getMemberId();
+		String textDefLanguageRefsetMemberId = referenceSetMemberService.findMembers("MAIN", textDefId, PAGE_OF_ONE).getContent().get(0).getMemberId();
+
+
+		// Test export
+		try (ZipInputStream zipInputStream = new ZipInputStream(new FileInputStream(exportFile))) {
+
+			ZipEntry axioms = zipInputStream.getNextEntry();
```
kaicode, line 520: Calling `getNextEntry` again after reading this one would prove that nothing else was exported. (Also this `axioms` variable has a misleading name.)

kaicode:
Closing because Deb raised a new one.

PR #267, opened 2021-05-17
RT2-56 Refset only export refactor / simplification.

kaicode:
@shapirod2 would this work as a further simplification to avoid code duplication? The unit tests you added still pass.


shapirod2 (approved):
OK.  I've also tested these revisions and they look good.  Thanks.

kaicode:
Merged, thanks @shapirod2!

PR #281, opened 2021-06-14
MAINT-1416 Restore correct details from parent code system

dmcgihtsdo:
MAINT-1416 is concerned with fixing a bug wherein the details of a Concept are not being reverted.

The bug will appear with this scenario: 

- Extension A creates a Concept et al
- Extension B modifies Extension A's data by inactivating a Relationship
- Extension B undoes the modifications made to Extension A's data
- The data modified (changed then unchanged) should be restored to its original state in Extension A. 

See the unit test: `ConceptUpdateHelperTest:saveNewOrUpdatedConcepts_ShouldRestoreDetailsFromParentCodeSystem`


dmcgihtsdo:
Some changes have been made to develop since I created this branch. I have rebased onto develop, hence the non MAINT-1416 commits.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -181,7 +180,6 @@ PersistedComponents saveNewOrUpdatedConcepts(
 				}
```
dmcgihtsdo, line 181: The main protagonist for this PR is `ConceptUpdateHelper` which has been tweaked to correctly restore the Concept's details.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -582,6 +581,17 @@ void doDeleteMembersWhereReferencedComponentDeleted(Set<String> entityVersionsDe
...
+			// Trying Concept module in attempt to restore effective time for the case
+			// where content has changed and then been reverted.
+			if (defaultModuleId != null) {
+				newComponent.setModuleId(newConcept.getModuleId());
+				newComponent.updateEffectiveTime();
+				if (newComponent.getEffectiveTime() == null) {
+					newComponent.setModuleId(defaultModuleId);
+				}
+			}
+
```
dmcgihtsdo, line 594: I added in the null-check as not all branches have a defaultModuleId. Without this check, a NullPointerException was eventually being thrown in various tests.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelperTest.java
```diff
@@ -0,0 +1,114 @@
...
+		// Extension B modifies Extension A's Concept by de-activating a Relationship
+		givenCodeSystemExists(SHORTNAME_AB, BRANCH_AB);
+		givenBranchExists(BRANCH_AB, MODULE_AB);
+		whenConceptModified(getConcept(BRANCH_A), false);
+		assertEffectiveTimeAndModuleId(getRelationship(getConcept(BRANCH_AB)), null, MODULE_AB);
+
+		// Extension B modifies Extension A's Concept (again) by reverting their change (inactivating Relationship)
+		whenConceptModified(getConcept(BRANCH_AB), true);
+		assertEffectiveTimeAndModuleId(getRelationship(getConcept(BRANCH_AB)), 20210131, MODULE_A); //Effective Time and Module Id restored from parent.
+	}
```
dmcgihtsdo, line 62: This unit test will fail without the appropriate code changes.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
pgwilliams: As discussed, the check for defaultModuleId here means that this code will only apply to MS branches so that avoids confusion with the Core / Model Module Ids.   It would be worth thinking about MS customers who might work in more than one module as we won't know which one to reset to.   I'll ask Andrew about that.

So you're saying here:    all components should take the moduleId of the parent concept, unless they've been updated in which case they should take the default module id.    I think I'm concerned about a case where a  model module concept has a description in the core module (which is wrong), but this code will automatically correct that which will make it look new and then pull the description over to the MS module.   I'll see if I can find an example of that for you.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelperTest.java
pgwilliams: GREAT unit test.   As we discussed, just to be belt and braces about it, I'd like you to target an existing unit test which will pass through your code and you do NOT want it to take effect.    So you've proven here that it does what you want, I think it would be good to also think about checking that it DOESN'T do what you DON'T want (if you'll excuse the double negative).    Why don't you see what you can find and I'll also check if we have any examples of components in an unexpected module which are then at risk.

pgwilliams:
> Some changes have been made to develop since I created this branch. I have rebased onto develop, hence the non MAINT-1416 commits.

Yes I'm surprised to see them.   When you rebased develop, did you then force push onto your feature branch to reset the "start" point there?   When I merged this one I think I'll try rebase and merge rather than a merge commit.

pgwilliams (approved):
Looks good @dmcgihtsdo thanks for walking through the changes with me and addressing my concerns about also checking that things we expect to NOT change do not get changed.   We also discussed the edge case of concepts with components in a different model which I think will be OK.

dmcgihtsdo:
> > Some changes have been made to develop since I created this branch. I have rebased onto develop, hence the non MAINT-1416 commits.
> 
> Yes I'm surprised to see them. When you rebased develop, did you then force push onto your feature branch to reset the "start" point there? When I merge this one I think I'll try rebase and merge rather than a merge commit.

@pgwilliams no, I just did a normal push. Should I have forced instead?

pgwilliams:
> > Yes I'm surprised to see them. When you rebased against develop, did you then force push onto your feature branch to reset the "start" point there? When I merge this one I think I'll try rebase and merge rather than a merge commit.
> @pgwilliams no, I just did a normal push. Should I have forced instead?
I believe so yes, because you're changing the point at which your previous commits are being added and therefore the checksums for your existing commits need to be changed.  In "git status" it will look unusual like "you're 11 commits ahead and 4 commits behind develop"   rather than just being purely ahead.

PR #292, opened 2021-07-03
Rt2 130b

RichardAWood:
Identifying whether children have descendants that are members of a specified refset.  Tried several different strategies: the current one is the best we've identified, but the performance is still not great.  Would appreciate any insight if there's a better approach we haven't stumbled on yet.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/QueryService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/QueryService.java
@@ -79,8 +79,6 @@ public class QueryService implements ApplicationContextAware {
 
 	private static final Function<Long, Object[]> CONCEPT_ID_SEARCH_AFTER_EXTRACTOR =
 			conceptId -> conceptId == null ? null : SearchAfterHelper.convertToTokenAndBack(new Object[]{conceptId});
-
-	private static Map<String, Page<Long>> refsetMemberCache = new HashMap<>();
 			
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
@@ -274,18 +272,6 @@ public class QueryService implements ApplicationContextAware {
 		Collection<Long> conceptIdFilter = null;
 		if (conceptQuery.conceptIds != null && !conceptQuery.conceptIds.isEmpty()) {
 			conceptIdFilter = conceptQuery.conceptIds.stream().map(Long::valueOf).collect(Collectors.toSet());
-		}
-		if(conceptQuery.getRefsetId() != null) {
-			final String refsetMemberQuery = "^ " + conceptQuery.getRefsetId();
-			if(refsetMemberCache.containsKey(refsetMemberQuery + "|" + branchCriteria + "|" + branchPath + "|" + conceptQuery.isStated())) {
-				conceptIdFilter = refsetMemberCache.get(refsetMemberQuery + "|" + branchCriteria + "|" + branchPath + "|" + conceptQuery.isStated()).getContent();
-			}
-			else {
-				Page<Long> refsetConceptIds = eclQueryService.selectConceptIds(refsetMemberQuery, branchCriteria, branchPath, conceptQuery.isStated(), null, null);
-				conceptIdFilter = refsetConceptIds.getContent();
-				refsetMemberCache.put(refsetMemberQuery + "|" + branchCriteria + "|" + branchPath + "|" + conceptQuery.isStated(), refsetConceptIds);
-			}
-
 		}
 		if (definitionStatusFilter != null || module != null) {
 			Page<Long> allConceptIds = eclQueryService.selectConceptIds(ecl, branchCriteria, branchPath, conceptQuery.isStated(), conceptIdFilter, null);
@@ -539,7 +525,6 @@ public class QueryService implements ApplicationContextAware {
 		private String ecl;
 		private Set<String> conceptIds;
 		private final DescriptionCriteria descriptionCriteria;
-		private String refsetId;
 
 		private ConceptQueryBuilder(boolean stated) {
 			this.stated = stated;
@@ -551,7 +536,7 @@ public class QueryService implements ApplicationContextAware {
 		}
 
 		private boolean hasLogicalConditions() {
-			return ecl != null || activeFilter != null || definitionStatusFilter != null || conceptIds != null || module != null || refsetId != null;
+			return ecl != null || activeFilter != null || definitionStatusFilter != null || conceptIds != null || module != null;
 		}
 
 		public ConceptQueryBuilder ecl(String ecl) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/QueryService.java
```diff
@@ -7,6 +7,7 @@
 import it.unimi.dsi.fastutil.longs.LongArraySet;
 import it.unimi.dsi.fastutil.longs.LongComparators;
 import org.elasticsearch.index.query.BoolQueryBuilder;
+import org.elasticsearch.search.sort.FieldSortBuilder;
```
kaicode, line 10: Please remove all the changes in this class, they are incorrect and not needed.

src/main/java/org/snomed/snowstorm/ecl/ECLQueryService.java
```diff
@@ -66,6 +67,17 @@
 		});
 	}
 
+	public Boolean hasAnyResults (String ecl, String branch, BranchCriteria branchCriteria, boolean stated, Collection<Long> conceptIdFilter) {
+		
+		SExpressionConstraint expressionConstraint = (SExpressionConstraint) eclQueryBuilder.createQuery(ecl);
+		Optional<Page<Long>> pageOptional = expressionConstraint.select(branch, branchCriteria, stated, conceptIdFilter, ControllerHelper.getPageRequest(0,1), queryService);
+		
+		if(pageOptional.get().hasNext()) {
```
kaicode, line 75: My understanding is that the `hasNext()` method will return true if there is another page of results after the first page. This means that if there is 1 query result the method will return false.

kaicode: It would be cleaner to return the result of this boolean expression in one line, rather than wrapping it in an `if` block and returning `true` or `false`.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -495,6 +500,25 @@ public AsyncConceptChangeBatch getConceptBulkChange(@PathVariable String branch,
...
+			// Calculate refset membership once, and pass in as a filter for each
+			// child-descendents query
+			Collection<Long> refsetMemberIds = eclQueryService.selectConceptIds("^" + refsetId,
+					versionControlHelper.getBranchCriteria(branch), branch, true, null, null).getContent();
+
+			for (ConceptMini child : children) {
+
+				if (eclQueryService.hasAnyResults("<< " + child.getConceptId(), branch,
+						versionControlHelper.getBranchCriteria(branch), true, refsetMemberIds)) {
+					child.addExtraField("descendantsAreMemberOfRefset", "true");
```
kaicode, line 515: Using a boolean value here, rather than a String, would be more consistent with the rest of the API.

Also you could avoid writing `child.addExtraField("descendantsAreMemberOfRefset",` on two separate lines by taking the boolean result of the `hasAnyResults` call and using that as the value.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -485,6 +489,7 @@ public AsyncConceptChangeBatch getConceptBulkChange(@PathVariable String branch,
 			@PathVariable String conceptId,
 			@RequestParam(defaultValue = "inferred") Relationship.CharacteristicType form,
 			@RequestParam(required = false, defaultValue = "false") Boolean includeDescendantCount,
+			@RequestParam(required = false, defaultValue = "") String refsetId,
```
kaicode, line 492: Optional parameters should default to `null`, not an empty string, for consistency with the rest of the app please.

kaicode: This parameter name does not seem explicit enough.. it triggers a behaviour where a flag is added to indicate further descendants within a refset.. perhaps this should be called `checkDescendantsWithinRefsetId` ? Please also add some notes to describe the behaviour of this parameter.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -495,6 +500,25 @@ public AsyncConceptChangeBatch getConceptBulkChange(@PathVariable String branch,
...
+		// refset
+		if (!refsetId.equals("")) {
+			// Calculate refset membership once, and pass in as a filter for each
+			// child-descendents query
+			Collection<Long> refsetMemberIds = eclQueryService.selectConceptIds("^" + refsetId,
+					versionControlHelper.getBranchCriteria(branch), branch, true, null, null).getContent();
+
+			for (ConceptMini child : children) {
+
+				if (eclQueryService.hasAnyResults("<< " + child.getConceptId(), branch,
```
kaicode, line 513: This is probably the most efficient way. I expect most of the time goes into fetching the refsetMemberIds above.
You could reuse the `BranchCriteria` created above rather than recreating it for every child.

kaicode (changes requested):
Thanks for this. I've left some feedback.

kaicode:
The only thing that might speed this up further would be to introduce an ECL cache for the results of the refset query, that is something I have been considering. The cache would need to expire when there is a new commit on that branch. Let me know if you think something like that is required. I should probably be the one to implement that.

src/main/java/org/snomed/snowstorm/core/data/services/QueryService.java
RichardAWood: This was left over from a previous change attempt - removed.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
RichardAWood: Agreed on both counts - implemented.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
RichardAWood: Ah, good point on the branchCriteria.   Change made.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
RichardAWood: Both changes made.

src/main/java/org/snomed/snowstorm/ecl/ECLQueryService.java
RichardAWood: Each return only ever includes one page due to "getPageRequest(0,1)", but agreed the hasNext is less than ideal.  Switching to hasContent(), and removing the if-block as described.

RichardAWood:
Closing out and continuing forward with RTT-299.

PR #296, opened 2021-07-15
FRI-123 Update classification PAC/SAC automatically

dmcgihtsdo:
FRI-123 is concerned with updating the PAC/SAC classification criteria on AAG whenever a commit is made. 

The majority of the functionality was written in FRI-61; this PR simply adds support for an edge case where the classification results are empty (as this does not trigger a commit).

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/classification/ClassificationService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/classification/ClassificationService.java
@@ -30,6 +30,7 @@ import org.snomed.snowstorm.core.data.repositories.classification.RelationshipCh
 import org.snomed.snowstorm.core.data.services.*;
 import org.snomed.snowstorm.core.data.services.classification.pojo.ClassificationStatusResponse;
 import org.snomed.snowstorm.core.data.services.classification.pojo.EquivalentConceptsResponse;
+import org.snomed.snowstorm.core.data.services.servicehook.CommitServiceHookClient;
 import org.snomed.snowstorm.core.pojo.LanguageDialect;
 import org.snomed.snowstorm.core.rf2.RF2Type;
 import org.snomed.snowstorm.core.rf2.export.ExportException;
@@ -109,6 +110,9 @@ public class ClassificationService {
 
 	@Autowired
 	private VersionControlHelper versionControlHelper;
+
+	@Autowired
+	private CommitServiceHookClient commitServiceHookClient;
 
 	@Autowired
 	private ConceptAttributeSortHelper conceptAttributeSortHelper;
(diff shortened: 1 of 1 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/BranchMetadataHelper.java
```diff
@@ -19,6 +19,8 @@
 	// Metadata for internal use only and can't be updated via branch metadata REST service
 	public static final String INTERNAL_METADATA_KEY = "internal";
 
+	public static final String CLASSIFIED_KEY = "classified";
```
dmcgihtsdo, line 22: I couldn't find a constant for classified, so I added it here beside what I also had to use.

src/main/java/org/snomed/snowstorm/core/data/services/classification/ClassificationService.java
```diff
@@ -242,6 +246,7 @@ private void init() throws ServiceException {
 												final boolean classified = !inferredRelationshipChangesFound && !equivalentConceptsFound;
 												BranchClassificationStatusService.setClassificationStatus(latestBranchCommit, classified);
 												branchService.updateMetadata(latestBranchCommit.getPath(), latestBranchCommit.getMetadata());
+												mockCommitCompletion(latestBranchCommit);
```
dmcgihtsdo, line 249: This method is triggered whenever a classification is completed.

src/main/java/org/snomed/snowstorm/core/data/services/classification/ClassificationService.java
```diff
@@ -438,6 +443,7 @@ public void saveClassificationResultsToBranch(String path, String classification
 				classification.setStatus(SAVED);
 			}
 
+			mockCommitCompletion(path, classification);
```
dmcgihtsdo, line 446: This method is triggered whenever a classification is updated manually.

kaicode (changes requested):
The branch classification status is already being set via `BranchClassificationStatusService.setClassificationStatus`. 
The only case I can think of where a classification status change is not already being sent to the AAG is when classification completes and there are no changes - in this situation the metadata is being updated but we need to post to the AAG. We already do this when creating a branch.

dmcgihtsdo:
> The branch classification status is already being set via `BranchClassificationStatusService.setClassificationStatus`.
> The only case I can think of where a classification status change is not already being sent to the AAG is when classification completes and there are no changes - in this situation the metadata is being updated but we need to post to the AAG. We already do this when creating a branch.

Sorry @kaicode , I don't quite understand. 

I think I'm already doing what you've said? For my testing, I've created a new branch, and then started a classification. The classification finishes with no Relationship changes and ends up at line 248 in ClassificationService. I am then sending a request to AAG with this information.

kaicode:
@dmcgihtsdo I am not suggesting what you have done will not work, I'm just suggesting that we could have more code reuse if you look at the other places where we are doing the same thing. I'll call you.

dmcgihtsdo:
> @dmcgihtsdo I am not suggesting what you have done will not work, I'm just suggesting that we could have more code reuse if you look at the other places where we are doing the same thing. I'll call you.

Thanks for the clarification Kai. 

I have removed the unnecessary call to the mock method (which also contained the redundant constant). Now all that is changed is the call to AAG when an empty classification.

kaicode (approved):
Looks good to me 👍

PR #297, opened 2021-07-20
RT2-126

RichardAWood:
This is adding a new method to lookup and return a list of concepts along with a single ancestor-path from the concepts to the top-level (not returning _all_ possible paths).

### DIFF
--- a/src/main/java/org/snomed/snowstorm/rest/ConceptController.java
+++ b/src/main/java/org/snomed/snowstorm/rest/ConceptController.java
@@ -576,7 +576,9 @@ public class ConceptController {
 		return findConceptsWithECL(">" + conceptId, form == Relationship.CharacteristicType.stated, branch, acceptLanguageHeader, 0, LARGE_PAGE.getPageSize()).getItems();
 	}
 
-	@GetMapping(value = "/browser/{branch}/concepts/ancestorPaths")
+	@ApiOperation(value = "Return concepts, and a path of each concept's ancestors to the top-level.",
+			notes = "Note: The output is intended to be displayed in a list view.  If there are multiple ancestor paths, only a single path will be returned per concept.")
+	@GetMapping(value = "/browser/{branch}/concepts/ancestor-paths")
 	@JsonView(value = View.Component.class)
 	public Collection<ConceptMini> findConceptAncestorPaths(@PathVariable String branch,
 			@RequestParam(required = false) List<Long> conceptIds,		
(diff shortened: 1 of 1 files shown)


src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -535,6 +576,44 @@ public AsyncConceptChangeBatch getConceptBulkChange(@PathVariable String branch,
...
+	}	
+	
+	private ArrayList<ConceptMini> ancestorPathHelper(String branch, Relationship.CharacteristicType form, String conceptId, ArrayList<ConceptMini> pathSoFar, String acceptLanguageHeader) {
+		Collection<ConceptMini> conceptParents = findConceptParents(branch, conceptId, form, false, acceptLanguageHeader);
+		
+		if(conceptParents.isEmpty()) {
+			return pathSoFar;
+		}
+		else {
+			ConceptMini conceptParent = conceptParents.iterator().next();
```
kaicode, line 610: A concept can have multiple parents. This method will randomly return "a" path, but not "the" path because in many cases "the" path can not be represented by a single array.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -535,6 +576,44 @@ public AsyncConceptChangeBatch getConceptBulkChange(@PathVariable String branch,
 		return findConceptsWithECL(">" + conceptId, form == Relationship.CharacteristicType.stated, branch, acceptLanguageHeader, 0, LARGE_PAGE.getPageSize()).getItems();
 	}
 
+	@GetMapping(value = "/browser/{branch}/concepts/ancestorPaths")
```
kaicode, line 579: Please add an `@ApiOperation` annotation with a `value` and some `notes` to explain what this does.

kaicode: Please change the mapping from `ancestorPaths` to `ancestor-paths`, we don't use camel case in the rest of the API.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
kaicode: If this is understood and acceptable for your application please make this limitation explicit in the API annotation notes (which become visible in the swagger page).

kaicode (changes requested):

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
RichardAWood: This was the intention - to choose a single path for the list view, rather than trying to calculate all possible paths (they can navigate up the other paths in the taxonomy view).

Adding notes to state this.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
RichardAWood: Changes made.

RichardAWood:
All requested changes made - pushing updates now.

kaicode (approved):
Looks good to me!

PR #298, opened 2021-07-20
FRI-54 Change CommitInformation to have source & target branch paths

dmcgihtsdo:
FRI-54 is concerned with either allowing or disallowing a project promotion depending on whether the appropriate criteria have been completed. The previous PR related to FRI-54 submitted was not fully tested and had a slight oversight.

Currently, when Snowstorm is promoting, the _target_ branch will be sent to AAG for criteria verification rather than the _source_ branch.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitInformation.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitInformation.java
@@ -7,21 +7,33 @@ import java.util.Map;
 
 public class CommitInformation {
 
-	private final String path;
+	private final String sourceBranchPath;
+	private final String targetBranchPath;
 	private final Commit.CommitType commitType;
 	private final long headTime;
 	private final Map<String, Object> metadata;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitInformation.java
```diff
@@ -7,21 +7,33 @@
...
+			this.sourceBranchPath = branch.getPath();
+			this.targetBranchPath = null;
+		} else {
+			this.sourceBranchPath = commit.getSourceBranchPath();
+			this.targetBranchPath = branch.getPath();
+		}
+	}
+
+	public String getSourceBranchPath() {
+		return sourceBranchPath;
```
dmcgihtsdo, line 32: The `sourceBranchPath` and `targetBranchPath` come from the same location for both `Rebase` and `Promotion`, however, are different when `Content` is the `CommitType`.

kaicode (approved):
Looks good! Nice catch! 😄

PR #300, opened 2021-07-22
RT2-153: Multisearch return refset buckets given descriptionId parameter

shapirod2:
Minor change to allow multisearch/referencesets to take descriptionId as a parameter.


kaicode (approved):
Looks good to me.

PR #302, opened 2021-08-04
Resolving NPE

RichardAWood:
As requested previously, changed the optional parameter to default to null, but had missed updating the .equals() call lower down.  Fixing.


kaicode (approved):

kaicode:
Great, thanks @RichardAWood

PR #303, opened 2021-08-06
FRI-54 Add another property to toggle whether promotions should be blocked

dmcgihtsdo:
FRI-54 is concerned with blocking promotion if the AAG indicates the associated criteria for a branch has not been completed. 

A flag has previously been added (`blockPromotion`) to toggle whether promotions should be blocked if the criteria has not been completed. This PR also adds another property (`failIfError`) to toggle whether promotions should be blocked if Snowstorm cannot communicate with AAG.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitServiceHookClient.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitServiceHookClient.java
@@ -1,6 +1,7 @@
 package org.snomed.snowstorm.core.data.services.servicehook;
 
 import io.kaicode.elasticvc.api.CommitListener;
+import io.kaicode.elasticvc.domain.Branch;
 import io.kaicode.elasticvc.domain.Commit;
 import org.ihtsdo.sso.integration.SecurityUtil;
 import org.slf4j.Logger;
@@ -11,6 +12,7 @@ import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Service;
 import org.springframework.util.StringUtils;
 import org.springframework.web.client.HttpClientErrorException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitServiceHookClient.java
```diff
@@ -24,24 +26,36 @@
 
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 	private final String serviceUrl;
+	private final boolean failIfError;
 	private final boolean blockPromotion;
 
 	public CommitServiceHookClient(@Value("${service-hook.commit.url}") String serviceUrl,
+								   @Value("${service-hook.commit.fail-if-error:true}") String failIfError,
```
dmcgihtsdo, line 33: By default, Snowstorm will block promotion if AAG cannot be reached.

src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitServiceHookClient.java
```diff
@@ -24,24 +26,36 @@
...
 								   @Value("${service-hook.commit.block-promotion-if-error:false}") String blockPromotion) {
 		this.serviceUrl = serviceUrl;
+		this.failIfError = Boolean.parseBoolean(failIfError);
 		this.blockPromotion = Boolean.parseBoolean(blockPromotion);
 		if (!StringUtils.isEmpty(serviceUrl)) {
 			final RestTemplateBuilder builder = new RestTemplateBuilder()
 					.rootUri(serviceUrl);
 			restTemplate = builder.build();
 		} else {
+			logger.info("CommitServiceHookClient is muted as service url not configured.");
```
dmcgihtsdo, line 43: Added some logging to show state of commit listener.

src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitServiceHookClient.java
```diff
@@ -51,35 +65,52 @@ public void preCommitCompletion(Commit commit) throws IllegalStateException {
...
+				logger.error("Cannot communicate with external system; throwing exception.");
+				throw new RuntimeServiceException("Cannot communicate with external system.", e);
+			} else {
+				logger.error("Cannot communicate with external system, however, failure will be ignored as per configuration.");
 			}
-
-			if (blockPromotion) {
-				logger.error("Promotion blocked; unexpected error.");
-				throw e;
+		} catch (HttpClientErrorException.Conflict e) {
```
kaicode, line 83: My understanding is that catch blocks are considered in top down order. I would be tempted to put this one as the first catch block because it's the most specific exception. The other catch which is less specific could continue to catch all instances of `RestClientException`, as long as it's below `HttpClientErrorException.Conflict`.

kaicode (changes requested):
Made one suggestion, looks good otherwise!

src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitServiceHookClient.java
dmcgihtsdo: Thanks Kai. 

I have made that change.

src/main/java/org/snomed/snowstorm/core/data/services/servicehook/CommitServiceHookClient.java
```diff
@@ -51,35 +65,52 @@ public void preCommitCompletion(Commit commit) throws IllegalStateException {
...
+				logger.info("Promotion blocked as not all criteria have been completed; throwing exception.");
+				throw new RuntimeServiceException("Promotion blocked as not all criteria have been completed.", e);
+			} else if (promotion) {
+				logger.info("Promotion will go ahead, despite not all criteria being completed, as per configuration.");
 			}
-
-			if (blockPromotion) {
-				logger.error("Promotion blocked; unexpected error.");
-				throw e;
+		} catch (ResourceAccessException e) {
```
kaicode, line 85: I think you should use a less specific exception here. Consider it's parent `RestClientException`. Otherwise we may experience unhandled exceptions.

dmcgihtsdo: Thanks Kai. 

Apologies for the silly mistake. I've made that change.

kaicode (approved):
Looks good to me.

PR #324, opened 2021-10-01
RT-259: Add 'leafFlagForm' to findConcepts call   /{branch}/concepts

shapirod2:


src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -175,6 +175,7 @@
 
 			@RequestParam(required = false) String ecl,
 			@RequestParam(required = false) String statedEcl,
+			@RequestParam(required = false) Relationship.CharacteristicType form,
```
kaicode, line 178: I'm a bit concerned that it's not clear that setting the "form" parameter will cause the endpoint to run slightly more slowly because of the additional fetch of the leaf flag. Since this is supposed to be the fast concept search endpoint (as opposed to the /browser one) I think requesting the leaf flag explicit.

- Please could you give "form" a default value of "inferred" please.
- Could you also add an "includeLeafFlag" boolean parameter that defaults to false and triggers the additional fetch when set to true.

Once this PR is accepted I will have a think about aligning the "ecl" and "statedEcl" parameters with the new "form" param.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -227,7 +228,12 @@
 		if (returnIdOnly) {
 			return new ItemsPage<>(queryService.searchForIds(queryBuilder, branch, pageRequest));
 		} else {
-			return new ItemsPage<>(queryService.search(queryBuilder, branch, pageRequest));
+			Page<ConceptMini> miniConcepts = queryService.search(queryBuilder, branch, pageRequest);
+			final BranchCriteria branchCriteria = versionControlHelper.getBranchCriteria(branch);
+			if (form != null && !form.equals("")) {
```
kaicode, line 233: form can not equal "" because it's an enum.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -227,7 +228,12 @@
 		if (returnIdOnly) {
 			return new ItemsPage<>(queryService.searchForIds(queryBuilder, branch, pageRequest));
 		} else {
-			return new ItemsPage<>(queryService.search(queryBuilder, branch, pageRequest));
+			Page<ConceptMini> miniConcepts = queryService.search(queryBuilder, branch, pageRequest);
+			final BranchCriteria branchCriteria = versionControlHelper.getBranchCriteria(branch);
```
kaicode, line 232: This should be within the if block.

src/main/java/org/snomed/snowstorm/rest/pojo/ConceptSearchRequest.java
```diff
@@ -1,149 +1,159 @@
...
-	public void setSearchAfter(String searchAfter) {
-		this.searchAfter = searchAfter;
-	}
-}
+package org.snomed.snowstorm.rest.pojo;
+
+import java.util.Set;
+import org.snomed.snowstorm.core.data.domain.Relationship;
+
+public class ConceptSearchRequest {
```
kaicode, line 6: Please only commit the lines you have changed.

kaicode (changes requested):

kaicode (approved):
Looks good, thanks for making those changes.

PR #328, opened 2021-10-06
FRI-127

dmcgihtsdo:
FRI-127 is concerned with returning certain criteria from authoring-acceptance-gateway whenever a snowstorm branch is flagged as having had content added via a batch process.

Snowstorm will set the appropriate flag whenever the import function is run or when updating/creating Concepts in bulk. This flag will later be picked up by authoring-acceptance-gateway.


src/main/java/org/snomed/snowstorm/core/data/services/SBranchService.java
```diff
@@ -194,4 +196,15 @@ public Date getPartialCommitTimestamp(String branchPath) {
...
+	public Branch setAuthorFlag(String branchPath, SetAuthorFlag setAuthorFlag) {
+		Branch branch = branchService.findBranchOrThrow(branchPath);
+
+		Metadata metadata = branch.getMetadata();
+		Map<String, String> authFlagMap = metadata.getMapOrCreate(AUTHOR_FLAGS_METADATA_KEY);
+		authFlagMap.put(setAuthorFlag.getName(), String.valueOf(setAuthorFlag.isValue()));
+		metadata.putMap(AUTHOR_FLAGS_METADATA_KEY, authFlagMap);
+
+		return branchService.updateMetadata(branchPath, metadata);
+	}
```
dmcgihtsdo, line 209: This logic was previously in `BranchController`, however I needed it in another place (`ConceptService`). 

I originally tried injecting the controller to the service, but that caused a circular dependency so I moved the logic to a separate service class.

src/main/java/org/snomed/snowstorm/core/rf2/rf2import/ImportService.java
```diff
@@ -169,6 +172,10 @@ private void setImportMetadata(RF2Type importType, String branchPath, boolean cr
 		if (importType == FULL || createCodeSystemVersion) {
 			internalMetadataMap.put(DISABLE_IMPORT_TRACEABILITY, "true");
 		}
+
+		Map<String, String> authFlagMap = metadata.getMapOrCreate(AUTHOR_FLAGS_METADATA_KEY);
+		authFlagMap.put(BATCH_CHANGE_KEY, "true");
+		metadata.putMap(AUTHOR_FLAGS_METADATA_KEY, authFlagMap);
```
dmcgihtsdo, line 178: When importing, the flag will be set on the branch.

PR #307, opened 2021-08-13
Designation use context FHIR extension implementation

danka74:
First attempt at implementing DUC FHIR extension: https://confluence.ihtsdotools.org/display/FHIR/Designation+extension
There are some things to discuss, e.g. how to turn this on or off or otherwise control the use of the extension.
Also, only added this for ValueSet expansion and not for CodeSystem lookup (but should be similar).

PR #333, opened 2021-10-28
Designation use context FHIR extension implementation

kaicode:
Raised on behalf of @danka74 (originally #307)

"First attempt at implementing DUC FHIR extension: https://confluence.ihtsdotools.org/display/FHIR/Designation+extension
There are some things to discuss, e.g. how to turn this on or off or otherwise control the use of the extension.
Also, only added this for ValueSet expansion and not for CodeSystem lookup (but should be similar)."


pgwilliams (approved):
Very clear code, good naming, structure and use of existing constants appreciated.

PR #340, opened 2021-11-15
FRI-27 Consume Classification status via JMS

dmcgihtsdo:
FRI-27 is concerned with updating Snowstorm to retrieve a Classification's status via JMS instead of HTTP.

This PR changes Snowstorm to request Classifications be written to a new queue and to listen to the new queue for Classification status updates.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/classification/JMSListenerClassificationStatusService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/classification/JMSListenerClassificationStatusService.java
@@ -1,0 +1,49 @@
+package org.snomed.snowstorm.core.data.services.classification;
+
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.activemq.command.ActiveMQTextMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.snomed.snowstorm.core.data.services.classification.pojo.ClassificationStatusResponse;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.jms.annotation.JmsListener;
+import org.springframework.stereotype.Component;
+
+import javax.jms.JMSException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+@ConditionalOnProperty(value = "classification-service.job.status.use-jms", havingValue = "true")
+public class JMSListenerClassificationStatusService implements ClassificationStatusService {
+	private static final Logger LOGGER = LoggerFactory.getLogger(JMSListenerClassificationStatusService.class);
+
+	private final ObjectMapper objectMapper = new ObjectMapper();
+	private final Map<String, ClassificationStatusResponse> classificationStatusChanges = Collections.synchronizedMap(new HashMap<>());
+
+	public JMSListenerClassificationStatusService(@Value("${classification-service.job.status.location}") String responseMessageQueue) {
+		if (responseMessageQueue == null || responseMessageQueue.isEmpty()) {
+			throw new IllegalStateException("No queue specified");
+		}
+	}
+
+	@JmsListener(destination = "${classification-service.job.status.location}")
+	void messageConsumer(ActiveMQTextMessage activeMQTextMessage) throws JMSException, JsonProcessingException {
+		try {
+			ClassificationStatusResponse response = objectMapper.readValue(activeMQTextMessage.getText(), ClassificationStatusResponse.class);
+			classificationStatusChanges.put(response.getId(), response);
+		} catch (JsonParseException | JsonMappingException e) {
+			LOGGER.error("Failed to parse message. Message: {}.", activeMQTextMessage.getText());
+		}
+	}
+
+	@Override
+	public ClassificationStatusResponse getStatusChange(String classificationId) {
+		return classificationStatusChanges.remove(classificationId);
+	}
+}
--- a/src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
@@ -1,11 +1,12 @@
 package org.snomed.snowstorm.core.data.services.classification;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.tomcat.util.http.fileupload.util.Streams;
 import org.elasticsearch.common.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.snomed.snowstorm.core.data.services.classification.pojo.ClassificationStatusResponse;
-import org.snomed.snowstorm.core.util.TimerUtil;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.core.io.FileSystemResource;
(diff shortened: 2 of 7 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
```diff
@@ -1,48 +1,75 @@
...
+import java.util.HashMap;
+import java.util.Map;
 
 @Service
 class RemoteClassificationServiceClient {
 
 	private final RestTemplate restTemplate;
-
 	private final Logger logger = LoggerFactory.getLogger(getClass());
+	private final ObjectMapper objectMapper = new ObjectMapper();
```
dmcgihtsdo, line 38: This will map `ActiveMQTextMessage` into our desired `ClassificationStatusResponse`.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
```diff
@@ -1,48 +1,75 @@
...
+	private final Map<String, ClassificationStatusResponse> classificationStatuses = new HashMap<>(); // ClassificationStatusResponse.id  => ClassificationStatusResponse
```
dmcgihtsdo, line 39: An internal cache to store all classification statuses retrieved from JMS/HTTP. After 1 our, the cache will be wiped.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
```diff
@@ -1,48 +1,75 @@
...
 
+	@JmsListener(destination = "${classification-service.job.destination-response}")
+	void messageConsumer(ActiveMQTextMessage activeMQTextMessage) throws JMSException, JsonProcessingException {
+		try {
+			ClassificationStatusResponse response = objectMapper.readValue(activeMQTextMessage.getText(), ClassificationStatusResponse.class);
+			classificationStatuses.put(response.getId(), response);
+		} catch (JsonParseException | JsonMappingException e) {
+			// ignore
+		}
+	}
```
dmcgihtsdo, line 66: Consume a message from the new status queue, wrapped in a try-catch in case anything ends up there that shouldn't have.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
```diff
@@ -1,48 +1,75 @@
...
+
+	@Scheduled(fixedRate = 3600000) // 1 hour
+	void clearClassificationStatus() {
+		classificationStatuses.clear();
+	}
+
```
dmcgihtsdo, line 72: After 1 hour, wipe the map containing the classification statuses.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
```diff
@@ -76,11 +104,15 @@ String createClassification(String previousPackage, String dependencyPackage, Fi
...
-		return classificationStatusResponse;
+		ClassificationStatusResponse response = classificationStatuses.get(classificationId);
+		if (response == null) {
+			response = restTemplate.getForObject("/classifications/{classificationId}", ClassificationStatusResponse.class, classificationId);
+			if (response != null) {
+				classificationStatuses.put(response.getId(), response);
+			}
+		}
+
+		return response;
```
dmcgihtsdo, line 115: When requesting the status of a classification, first check the internal map. If it is not there, then fall back to requesting directly from the classification service.

src/main/resources/application.properties
```diff
@@ -303,6 +303,8 @@ classification-service.password=
 # Classification Job Timeout in minutes
 classification-service.job.abort-after-minutes=45
 
+# Queue containing status of classification
+classification-service.job.destination-response=classification-jobs-status
```
dmcgihtsdo, line 307: This queue is where Snowstorm requests Classification Service to write a response to after processing a classification. I suspect this value will change per environment.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
```diff
@@ -1,48 +1,75 @@
...
 				.build();
 	}
 
+	@JmsListener(destination = "${classification-service.job.destination-response}")
+	void messageConsumer(ActiveMQTextMessage activeMQTextMessage) throws JMSException, JsonProcessingException {
+		try {
+			ClassificationStatusResponse response = objectMapper.readValue(activeMQTextMessage.getText(), ClassificationStatusResponse.class);
+			classificationStatuses.put(response.getId(), response);
+		} catch (JsonParseException | JsonMappingException e) {
+			// ignore
```
kaicode, line 64: Please log an error.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
kaicode: I suggest we rename this method to `getStatusChange` and return null if there is no new status delivered via JMS. 

The `classificationStatuses` collection could be renamed to `classificationStatusChanges` and the status removed when `getStatusChange` is called. There would be no need to clear the collection on a scheduled basis. `ClassificationService` would need to be updated to deal with status changes and expect null values.

src/main/resources/application.properties
kaicode: We need to make a breaking change note in the release notes about this because people will now need a JMS broker set up to be able to use classification. Either that or we make this backwards compatible with a config option to switch between the two strategies. If the status update strategy is polling then `getStatusChange` would always fetch the latest status? The `ClassificationService` class would have to check if the status is null and if it's new or not..

dmcgihtsdo: Great suggestion, thanks Kai. 

I have changed this by adding `JMSListenerClassificationStatusService` and `RestTemplateClassificationStatusService` which can be controlled via the property `classification-service.job.status.use-jms`.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RemoteClassificationServiceClient.java
dmcgihtsdo: Thanks, I have done this.

src/main/java/org/snomed/snowstorm/core/data/services/classification/RestTemplateClassificationStatusService.java
```diff
@@ -0,0 +1,34 @@
...
+
+	private final RestTemplate restTemplate;
+
+	public RestTemplateClassificationStatusService(@Value("${classification-service.url}") String serviceUrl,
+												   @Value("${classification-service.username}") String serviceUsername,
+												   @Value("${classification-service.password}") String servicePassword,
+												   @Value("${classification-service.job.status.location}") String responseMessageQueue) {
+		if (responseMessageQueue != null && !responseMessageQueue.isEmpty()) {
+			throw new IllegalStateException("Queue specified but not required for this strategy.");
+		}
```
dmcgihtsdo, line 22: If there is a conflict in properties, Snowstorm will fail to start.

src/main/resources/application.properties
```diff
@@ -303,6 +303,11 @@ classification-service.password=
 # Classification Job Timeout in minutes
 classification-service.job.abort-after-minutes=45
 
+# Strategy to retrieve status of a classification. False by default for backward compatibility.
+classification-service.job.status.use-jms=false
+
+# Queue containing the status of a classification. Blank by default for backward compatibility.
+classification-service.job.status.location=
```
dmcgihtsdo, line 310: If this approach is approved, I'll log an INFRA ticket for our environments to use JMS. By default, Snowstorm will not use JMS.

src/main/java/org/snomed/snowstorm/core/data/services/classification/JMSListenerClassificationStatusService.java
```diff
@@ -0,0 +1,53 @@
...
+		try {
+			ClassificationStatusResponse response = objectMapper.readValue(activeMQTextMessage.getText(), ClassificationStatusResponse.class);
+			classificationStatusChanges.put(response.getId(), response);
+		} catch (JsonParseException | JsonMappingException e) {
+			LOGGER.error("Failed to parse message. Message: {}.", activeMQTextMessage.getText());
+		}
+	}
+
+	@Override
+	public ClassificationStatusResponse getStatusChange(String classificationId) {
```
kaicode, line 45: The content of this method could simply be:
```
return classificationStatusChanges.remove(classificationId);
```
?

src/main/java/org/snomed/snowstorm/core/data/services/classification/JMSListenerClassificationStatusService.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+@ConditionalOnProperty(value = "classification-service.job.status.use-jms", havingValue = "true")
+public class JMSListenerClassificationStatusService implements ClassificationStatusService {
+	private static final Logger LOGGER = LoggerFactory.getLogger(JMSListenerClassificationStatusService.class);
+
+	private final ObjectMapper objectMapper = new ObjectMapper();
+	private final Map<String, ClassificationStatusResponse> classificationStatusChanges = new HashMap<>();
```
kaicode, line 26: I would use some sort of synchronised map here to avoid the slim possibility that one thread adds something in the same moment as another removing the latest response. Not that slim seeing as there is a thread that pulls statuses out every second.

kaicode (approved):
Great improvement. Much simpler. Thank you for making this backwards compatible.
I've made a couple of minor suggestions but approving this anyway.

PR #345, opened 2021-12-06
Fixes #344 Upgrade test libs and jna to fix unit tests on arm64.

### LINKED ISSUE #344: Unit tests do not run on MacOS arm64 architecture
Building the project with unit tests fails on a Mac with M1 chip. The Docker container for Elasticsearch can not launch.

kaicode:
Fixes build for Mac M1 chip.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,12 @@
 			<groupId>io.kaicode</groupId>
 			<artifactId>elasticvc</artifactId>
 			<version>${elasticvc.version}</version>
+			<exclusions>
+				<exclusion>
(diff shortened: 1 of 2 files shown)


kaicode:
As well as upgrading the testcontainer and JNA libraries I had to get unit tests to pull the container for Elasticsearch 7.10.0 too. The version we were using (7.7.0) would not run because it was using an old JNA version that doesn't work with arm64.

CoderMChu (approved):
@kaicode  It looks good. Thanks.

PR #349, opened 2021-12-13
FRI-234 Auto merge source content if source versioned and target deleted

dmcgihtsdo:
FRI-234 is concerned with updating merge to cover scenarios where content is deleted after it has been versioned.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
@@ -163,6 +163,11 @@ public class BranchReviewService {
 					persistManualMergeConceptDeletion(mergeReview, conceptId);
 					logger.info("Concept {} deleted on both sides of the merge. Excluding from merge review {}.", conceptId, id);
 				} else {
+					if (sourceVersion != null && sourceVersion.isReleased() && targetVersion == null) {
+						// Deleted somewhere, whilst simultaneously versioned elsewhere.
+						mergeVersion.setAutoMergedConcept(sourceVersion);
+					}
+
 					if (sourceVersion != null && targetVersion != null) {
 						// Neither deleted, auto-merge.
 						mergeVersion.setAutoMergedConcept(autoMergeConcept(sourceVersion, targetVersion));
@@ -542,7 +547,7 @@ public class BranchReviewService {
 		NativeSearchQueryBuilder builder = new NativeSearchQueryBuilder()
 				.withQuery(boolQuery()
 						.must(termsQuery("_id", versionsReplaced))
-						.mustNot(getNonStatedRelationshipClause()))
+						.mustNot(getNonStatedModifiedReleasedRelationshipClause()))
 				.withPageable(LARGE_PAGE);
 		if (limitFieldsFetched.length > 0) {
 			builder.withFields(limitFieldsFetched);
@@ -556,6 +561,13 @@ public class BranchReviewService {
 						.must(branchUpdatesCriteria)
 						.mustNot(getNonStatedRelationshipClause()))
 				.withPageable(LARGE_PAGE);
+	}
+
+	private QueryBuilder getNonStatedModifiedReleasedRelationshipClause() {
+		return boolQuery()
+				.must(existsQuery(Relationship.Fields.EFFECTIVE_TIME))
+				.mustNot(existsQuery(Relationship.Fields.RELEASED))
+				.mustNot(termsQuery(Relationship.Fields.CHARACTERISTIC_TYPE_ID, Concepts.INFERRED_RELATIONSHIP, Concepts.ADDITIONAL_RELATIONSHIP));
 	}
 
 	private QueryBuilder getNonStatedRelationshipClause() {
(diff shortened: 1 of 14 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -163,6 +163,11 @@ public MergeReview getMergeReviewOrThrow(String id) {
 					persistManualMergeConceptDeletion(mergeReview, conceptId);
 					logger.info("Concept {} deleted on both sides of the merge. Excluding from merge review {}.", conceptId, id);
 				} else {
+					if (sourceVersion != null && sourceVersion.isReleased() && targetVersion == null) {
+						// Deleted somewhere, whilst simultaneously versioned elsewhere.
+						mergeVersion.setAutoMergedConcept(sourceVersion);
+					}
+
```
dmcgihtsdo, line 170: This block is essentially the fix. 

If the source is released and the target is null (it was deleted), then auto merge the source.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -558,6 +563,13 @@ private NativeSearchQueryBuilder newSearchQuery(BoolQueryBuilder branchUpdatesCr
 				.withPageable(LARGE_PAGE);
 	}
 
+	private QueryBuilder getNonStatedModifiedReleasedRelationshipClause() {
+		return boolQuery()
+				.must(existsQuery(Relationship.Fields.EFFECTIVE_TIME))
+				.mustNot(existsQuery(Relationship.Fields.RELEASED))
+				.mustNot(termsQuery(Relationship.Fields.CHARACTERISTIC_TYPE_ID, Concepts.INFERRED_RELATIONSHIP, Concepts.ADDITIONAL_RELATIONSHIP));
+	}
+
```
dmcgihtsdo, line 572: When writing a unit test, I learned that my Relationships (inferred relationships to mimic released) were not being found in the merge. I dug through and found it was from the `getNonStatedRelationshipClause` method explicitly ignoring inferred Relationships. 

I have therefore created `getNonStatedModifiedReleasedRelationshipClause` to limit the scope of my change, but I think this is correct: find Relationships that are released, but also modified.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -1834,6 +1834,144 @@ void testPromotionAbortedIfBranchReviewIncomplete() throws ServiceException {
 		assertEquals("Promotion blocked; not all criteria have been met.", message);
 	}
 
+	@Test
+	void testAutoMergeWhenConceptDeletedSomewhereAndVersionedElsewhere() throws ServiceException, InterruptedException {
```
dmcgihtsdo, line 1838: This tests Concept deletion post versioning.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -1834,6 +1834,144 @@ void testPromotionAbortedIfBranchReviewIncomplete() throws ServiceException {
...
+		assertEquals(concept, conflicts.iterator().next().getAutoMergedConcept());
+
+		reviewService.persistManualMergeConceptDeletion(reviewService.getMergeReviewOrThrow(review.getId()), Long.parseLong(concept.getId()));
+		reviewService.applyMergeReview(review);
+
+		assertNotNull(conceptService.find(concept.getId(), childBranch)); // Confirm restored post rebase
+	}
+
+	@Test
+	void testAutoMergeWhenDescriptionDeletedSomewhereAndVersionedElsewhere() throws ServiceException, InterruptedException {
```
dmcgihtsdo, line 1884: This tests Description deletion post versioning.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -1834,6 +1834,144 @@ void testPromotionAbortedIfBranchReviewIncomplete() throws ServiceException {
...
+		assertEquals(concept, conflicts.iterator().next().getAutoMergedConcept());
+
+		reviewService.persistManualMergeConceptDeletion(reviewService.getMergeReviewOrThrow(review.getId()), Long.parseLong(concept.getId()));
+		reviewService.applyMergeReview(review);
+
+		assertNotNull(descriptionService.findDescription(childBranch, syn.getId())); // Confirm restored post rebase
+	}
+
+	@Test
+	void testAutoMergeWhenRelationshipNormalDeletedSomewhereAndVersionedElsewhere() throws ServiceException, InterruptedException {
```
dmcgihtsdo, line 1929: This tests Relationship deletion post versioning.

kaicode:
The "automerged" concept is a conflict resolution suggestion. It is presented to the user in the middle column of the authoring platform conflict screen. This PR seems to make the automerged concept the released concept as a suggestion.

The implementation of this ticket should:
- Prevent released concepts from being deleted
- Prevent released components from being deleted
This should happen during applying the merge review rather than in the step that makes suggestions about merge resolution. I would look at `BranchReviewService.applyMergeReview()` and assume the user has picked the right hand side of the merge where the concept is delete or some description is deleted when the right hand side is versioned.

dmcgihtsdo:
Thanks for the feedback @kaicode .

I'll reopen a new PR with those changes.

PR #368, opened 2022-01-31
Fixes #367 return concept ids as strings when returnIdOnly param used.

### LINKED ISSUE #367: Concept search "returnIdOnly" format must use quoted ids to avoid javascript precision issue
Concept ids that are 18 characters long get rounded when put into JSON as a number.
For example "450229981000132104" is rounded to "450229981000132100".. the last two digits get changed to 0.

Snowstorm should always return SCTIDs as quoted Strings to avoid this problem. The concept search is not currently doing this when the `returnIdOnly` parameter is used. Running a search that matches concept "450229981000132104" returns an identifier of "450229981000132104" in the response. This is a bug.

https://snowstorm.ihtsdotools.org/snowstorm/snomed-ct/MAIN/concepts?conceptIds=450229981000132104&returnIdOnly=true
Screenshot 2022-01-31 at 15 37 39

kaicode:
Controller unit tests still pass.
Local manual tests look good:
Screenshot 2022-01-31 at 16 18 19

Search after still works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 7.7.0 Release
+
+### Breaking
+- General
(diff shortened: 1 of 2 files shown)


pgwilliams (approved):
Looks great.

PR #376, opened 2022-02-10
Rollup delta warnings

kaicode:
Added warnings about rollup deltas to the general Loading SNOMED doc.
Updated advice on the extension loading doc about monthly releases and using snapshots. Also updated dates in examples.


rorydavidson (approved):
Looks good, thanks.

PR #377, opened 2022-02-11
Add property latestDailyBuild

hurtigcodes:
Aim of the pull request is to make it more obvious for users how to find the latest db. Current ways of inspecting branches or checking authoring stats doesn't give a simple answer to the questions: has the db been loaded today, and if so, which version has Snowstorm loaded?

Two things here:
1. Added a property to codesystems: latestDailyBuild
2. A small endpoint which only checks latestDailyBuild against today's date. Obviously, this isn't a 'necessary' endpoint to add. The rationale behind it was to move that logic from the client to Snowstorm to enable users to use simpler alerts/notification tools (or for non-technical users who cannot define that logic themselves) for triggering a failed db alert.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/domain/CodeSystem.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/domain/CodeSystem.java
@@ -20,7 +20,7 @@ import java.util.Set;
  * Represents an edition or extension of SNOMED-CT
  */
 @Document(indexName = "codesystem")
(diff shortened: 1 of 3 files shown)


kaicode (approved):
Looks good to me.

kaicode:
@CoderMChu Looks okay to me. I'll leave it with you. Could change the merge target to develop.

CoderMChu:
@hurtigcodes Thanks for the pull request.

PR #378, opened 2022-02-21
MAINT-1882 Snowstorm IPS Import without MRCM

kaicode:
From ticket:
> We would like to exclude the MRCM refsets from the IPS terminology.
>
> The MRCM refsets are currently required by the ECL index service within Snowstorm to check the data type of the concrete values being indexed.
>
> Try to update the Snowstorm import and ECL index service to work without the MRCM by guessing the data type. All numbers could be indexed as decimals because "=#1" and "=#1.0" are treated the same in ECL anyway.
> 
> This will allow us to exclude the MRCM refsets from the IPS and other future subontologies / terminology products.

This pull request stops the RF2 import from failing when the MRCM is missing. Number concrete domains will default to decimal data type if missing from the MRCM. ECL requests still work as expected as shown in unit tests.


rorydavidson:
Nice, thanks @kaicode

kaicode:
Thanks @CoderMChu

Issue #379, opened 2022-03-11
Wrong PT is provided by browser end point


danka74:
For the concept 252616000 | Auditory brainstem electric response audiometry (procedure) | snowstorm provides the wrong PT in Swedish.

https://snowstorm.ihtsdotools.org/snowstorm/snomed-ct/browser/MAIN/SNOMEDCT-SE/2021-11-30/concepts/252616000?descendantCountForm=inferred

```
{
conceptId: "252616000",
descendantCount: 8,
fsn: {
term: "Auditory brainstem electric response audiometry (procedure)",
lang: "en"
},
pt: {
term: "ABR",
lang: "sv"
},
active: true,
effectiveTime: "20020131",
released: true,
releasedEffectiveTime: 20020131,
moduleId: "900000000000207008",
definitionStatus: "PRIMITIVE",
descriptions: [
...
{
active: true,
moduleId: "45991000052106",
released: true,
releasedEffectiveTime: 20210531,
descriptionId: "4795141000052117",
term: "ABR",
conceptId: "252616000",
typeId: "900000000000013009",
acceptabilityMap: {
46011000052107: "ACCEPTABLE"
},
type: "SYNONYM",
caseSignificance: "ENTIRE_TERM_CASE_SENSITIVE",
lang: "sv",
effectiveTime: "20210531"
},
...
```

kaicode:
Hi @danka74,

I am able to reproduce this issue when using request header `Accept-Language: sv`. This causes the term "" to be returned, which is incorrect because this term is only acceptable, not preferred.

The API seems to behave correctly when a specific language refset is requested using header `Accept-Language: sv-X-46011000052107`.

danka74:
Hi @kaicode ,
ok, the browser does not use the language refset specification in Accept-Language it seems, so in the browser the wrong term is displayed.
/Daniel

kaicode:
@danka74 thanks for pointing this out. We will get this bug fixed.

kaicode:
Jira ticket - https://jira.ihtsdotools.org/browse/BROWSE-531

danka74:
Thanks!

danka74:
Still seems to be an issue.
http://localhost:8080/snowstorm/snomed-ct/browser/MAIN%2FSNOMEDCT-SE/concepts/122575003
(running develop branch 7.8.0-SNAPSHOT)
[...]

PR #381, opened 2022-03-12
default lang refset for lang only accept header

danka74:
When providing just a 2-character language tag with Accept-language (no refset or dialect), the language reference set selected (and the PT provided by snowstorm) becomes slightly random. The PT provided will be any preferred term in any lanugage reference set in the order they are provided.

By adding the default language reference set under "# SNOMED Known Dialects - Individual Configuration" in the application.properties file and adding this default lanugage reference set when a 2-character Accept-language header is given, the assigning of PT becomes predictable.


danka74:
Fixes #379

kaicode:
Looks good to me.

PR #385, opened 2022-03-28
FRI-210 Preserve International changes after Extension upgrade

dmcgihtsdo:
FRI-210 is concerned with preventing International changes being lost whenever an Extension does an upgrade/a rebase.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
@@ -170,6 +170,9 @@ public class BranchReviewService {
 					if (sourceVersion != null && targetVersion != null) {
 						// Neither deleted, auto-merge.
 						mergeVersion.setAutoMergedConcept(autoMergeConcept(sourceVersion, targetVersion));
+
+						// Upgrade components if behind a release/version
+						mergeVersion.setTargetConcept(upgradeComponents(sourceVersion, targetVersion));
 					}
 					conflicts.put(conceptId, mergeVersion);
 				}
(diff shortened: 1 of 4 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -170,6 +170,9 @@ public MergeReview getMergeReviewOrThrow(String id) {
 					if (sourceVersion != null && targetVersion != null) {
 						// Neither deleted, auto-merge.
 						mergeVersion.setAutoMergedConcept(autoMergeConcept(sourceVersion, targetVersion));
+
+						// Upgrade components if behind a release/version
+						mergeVersion.setTargetConcept(upgradeComponents(sourceVersion, targetVersion));
```
dmcgihtsdo, line 175: This block sets the RHS of the merge to be an "upgraded" version.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -228,6 +231,106 @@ public void applyMergeReview(MergeReview mergeReview) throws ServiceException {
 		branchMergeService.mergeBranchSync(mergeReview.getSourcePath(), mergeReview.getTargetPath(), concepts);
 	}
 
+	private Concept upgradeComponents(Concept sourceVersion, Concept targetVersion) {
```
dmcgihtsdo, line 234: Essentially, this method copies properties from the latest Component (either Source or Target version). If International has been upgraded, then the Source version will be used.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -255,6 +358,9 @@ private Concept autoMergeConcept(Concept sourceConcept, Concept targetConcept) {
 		mergedConcept.setDefinitionStatus(winningConcept.getDefinitionStatus());
 		mergedConcept.setEffectiveTimeI(winningConcept.getEffectiveTimeI());
 		mergedConcept.setModuleId(winningConcept.getModuleId());
+		mergedConcept.setReleased(winningConcept.isReleased());
+		mergedConcept.setReleaseHash(winningConcept.getReleaseHash());
+		mergedConcept.setReleasedEffectiveTime(winningConcept.getReleasedEffectiveTime());
```
dmcgihtsdo, line 363: I noticed these properties weren't copied over when I was stepping through some tests.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -606,7 +606,7 @@ private void doSaveBatchQueryConcept(Collection<QueryConcept> queryConcepts, Com
 			
 			
 			//Any change to a component in an extension needs to be done in the default module
-			if (newComponent.isChanged() && (defaultModuleId != null && 
+			if (!rebase && newComponent.isChanged() && (defaultModuleId != null &&
 					!defaultModuleId.equals(Concepts.CORE_MODULE))) {
 				newComponent.setModuleId(defaultModuleId);
 			}
```
dmcgihtsdo, line 612: `newComponent` will be flagged as `changed` whenever doing a rebase. As a result, `newComponent` is having its module changed on a rebase. This caused Descriptions to lose their effectiveTime / move module for one of my tests.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -2109,6 +2116,824 @@ void testAutoMergeWhenMemberDeletedSomewhereAndVersionedElsewhere() throws Servi
 		assertVersioned(referenceSetMember, 20210812);
 	}
 
+	/*
+	 * The Extension upgrades to the latest version of International. After doing the upgrade, the Extension's task then
+	 * does a rebase, which results in no conflicts as the task has no authoring changes. The latest and greatest components
+	 * are present on the task.
+	 * */
+	@Test
+	void testUpgradeExtensionWhenExtensionTaskHasNoAuthoringChanges() throws ServiceException, InterruptedException {
```
dmcgihtsdo, line 2125: A number of tests have been added which cover our 4 scenarios we spoke about privately @CoderMChu

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -228,6 +231,106 @@ public void applyMergeReview(MergeReview mergeReview) throws ServiceException {
...
+				mergedAxioms.put(targetAxiomId, targetAxiom);
+			} else {
+				ReferenceSetMember sourceMember = sourceAxiom.getReferenceSetMember();
+				ReferenceSetMember targetMember = targetAxiom.getReferenceSetMember();
+
+				if (!sourceMember.isReleasedMoreRecentlyThan(targetMember)) {
+					mergedAxioms.put(targetAxiomId, targetAxiom);
+				}
+			}
+		}
```
CoderMChu, line 329: @dmcgihtsdo  We can create one method that uses snomed components as arguments to do merge for descriptions, relationships and axioms.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
CoderMChu: Good catch. I thought we are going to improve the auto merge logic too in our discussions.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
CoderMChu: Can you please add a test case for this change?

CoderMChu (changes requested):
@dmcgihtsdo We can have a quick discussion tomorrow about the comments I have added.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
dmcgihtsdo: If I remove `!rebase` or `newComponent.isComponentChanged(existingComponent)` then two tests fail: 

`testUpgradeExtensionWhenExtensionProjectAndTaskBothHaveAuthoredSameComponentAndUserSelectsMiddle` and `testUpgradeExtensionWhenExtensionProjectAndTaskBothHaveAuthoredSameComponentAndUserSelectsRHS`. 

These fails because the `effectiveTime` and `module` have incorrectly changed for some Components.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -228,6 +231,81 @@ public void applyMergeReview(MergeReview mergeReview) throws ServiceException {
...
+
+		return upgradedConcept;
+	}
+
+	private <T extends SnomedComponent> Set<T> mergePublished(Set<T> sourceComponents, Set<T> targetComponents) {
+		Map<String, T> mergedComponents = new HashMap<>();
+
+		for (T sourceComponent : sourceComponents) {
+			boolean publishedAndUnchanged = sourceComponent.getReleasedEffectiveTime() != null && sourceComponent.getEffectiveTimeI() != null;
+			if (publishedAndUnchanged) {
```
CoderMChu, line 262: sourceComponent.getEffectiveTimeI() != null is sufficient

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -228,6 +231,81 @@ public void applyMergeReview(MergeReview mergeReview) throws ServiceException {
...
+				mergedComponents.put(sourceComponent.getId(), sourceComponent);
+			}
+		}
+
+		for (T targetComponent : targetComponents) {
+			String targetDescriptionId = targetComponent.getId();
```
CoderMChu, line 268: targetComponentId is better name here.

CoderMChu (commented):
It looks better now. Thanks

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
CoderMChu: After further checking the logic here is not quite right. We should only replace the target with the source version when both versions have effective time and the source version is newer than the target. I have drafted the logic and will send it to you.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -256,22 +256,30 @@ private Concept upgradeComponents(Concept sourceVersion, Concept targetVersion)
 
 	private <T extends SnomedComponent> Set<T> mergePublished(Set<T> sourceComponents, Set<T> targetComponents) {
 		Map<String, T> mergedComponents = new HashMap<>();
-
 		for (T sourceComponent : sourceComponents) {
-			boolean publishedAndUnchanged = sourceComponent.getReleasedEffectiveTime() != null && sourceComponent.getEffectiveTimeI() != null;
-			if (publishedAndUnchanged) {
+			if (sourceComponent.getEffectiveTimeI() != null) {
```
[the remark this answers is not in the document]
dmcgihtsdo: @CoderMChu If I do not exclude unpublished components, then several tests fail because the task will receive the project's changes despite the author picking the RHS. 

- testUpgradeExtensionWhenExtensionProjectAndTaskBothHaveAuthoringChangesAndUserSelectsRHS
- testChangingComponentAfterPublishingAndUserSelectsRHS
- testUpgradeExtensionWhenExtensionProjectAndTaskBothHaveAuthoredSameComponentAndUserSelectsRHS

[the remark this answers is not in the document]
CoderMChu: @dmcgihtsdo  I had a look at the first test case above. The failed assertion doesn't look be correct. Let's have a quick call.

PR #387, opened 2022-03-29
MAINT-1743 - Changed markDeletionsAndUpdates(), added test cases

astro-snail:
Also includes change to the same method from the branch feature/FRI-210


src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1686,6 +1686,109 @@ public void testBulkJobAxiomSerialisationError() {
...
+
+		// 9. Revert the change in Step 7. (i.e change the caseSignificanceId back to what was before) and save.
+		description.setCaseSignificanceId(CASE_INSENSITIVE);
+		conceptService.update(concept, "MAIN/SNOMEDCT-BE/BE/BE-1");
+
+		// 10. The effective time should NOT be cleared.
+		concept = conceptService.find("100001", "MAIN/SNOMEDCT-BE/BE/BE-1");
+		description = concept.getDescription("100001");
+		assertEquals(CASE_INSENSITIVE, description.getCaseSignificanceId());
+		assertEquals("11000172109", description.getModuleId());
```
pgwilliams, line 1784: Yes, this is what I was in particularly looking for.   Great!

pgwilliams (approved):
Really great.   Love the comments on each step of the unit test, very clear.

I was thinking of a use case where an extension adds langrefset entries to a core **en** description, but however we do that, any change on an extension would appear in the extension default module so I think we've got that covered.

I think we've got more use cases to create unit tests for until we're water tight, but as far as this ticket goes, it looks really good.   Approved.

astro-snail:
Thank you, @pgwilliams, but all credit regarding steps and comments goes to @CoderMChu 
I just kept them and expanded on them a bit in the second test case. Agree that we need more use cases.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -598,28 +598,27 @@ private void doSaveBatchQueryConcept(Collection<QueryConcept> queryConcepts, Com
...
+
+			// Any change to a component in an extension needs to be done in the default module
+			if (newComponent.isComponentChanged(existingComponent) && (defaultModuleId != null && !defaultModuleId.equals(Concepts.CORE_MODULE))) {
 				newComponent.setModuleId(defaultModuleId);
 			}
+			newComponent.updateEffectiveTime();
 
 			// Trying Concept module in attempt to restore effective time for the case
 			// where content has changed and then been reverted.
-			if (shouldRestoreDefaultModuleId(defaultModuleId, existingComponent, newComponent)) {
```
CoderMChu, line 616: @astro-snail  This fix seems to have fixed your bug but there are other cases which are still not correct. e.g new components created on international are having null module id. I have listed test scenarios in a google sheet and will share with you.

pgwilliams:
Where are we with this code please @astro-snail ?   Do you think it's ready to merge?   Are you happy with the unit tests?  Thanks!

astro-snail:
> Where are we with this code please @astro-snail ? Do you think it's ready to merge? Are you happy with the unit tests? Thanks!
@pgwilliams no, it is not ready to merge. I made the failing test pass locally, but I have not pushed that change to remote, as I believe it is not in a good shape yet. I was advised to put it on hold for a while and work on MAINT-1644.

astro-snail:
Not ready for review yet.

PR #391, opened 2022-04-07
FRI-424 - Changed startNewAuthoringCycle() method; updated tests

astro-snail:
@QuyenLy87 and @pgwilliams adding you as reviewers in case you have time to review before @CoderMChu.
Thanks!


src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
```diff
@@ -577,14 +577,28 @@ public void updateCodeSystemBranchMetadata(CodeSystem codeSystem, CodeSystemNewA
...
+			// Get previous package
+			final CodeSystemVersion codeSystemVersion = findVersion(codeSystem.getShortName(), Integer.parseInt(updateRequest.getPreviousRelease()));
+			if (codeSystemVersion == null) {
+				throw new IllegalStateException("Previous release " + updateRequest.getPreviousRelease() + " not found.");
+			}
+			branchMetadata.putString(PREVIOUS_PACKAGE, codeSystemVersion.getReleasePackage());
 		}
-		if (updateRequest.getPreviousDependency() != null) {
-			branchMetadata.putString(PREVIOUS_DEPENDENCY_PACKAGE, updateRequest.getPreviousDependency());
+		if (updateRequest.getPreviousDependencyRelease() != null && codeSystem.getDependantVersionEffectiveTime() != null) {
```
CoderMChu, line 590: @astro-snail  It looks good. After checking it further, we can derive previousDependencyRelease from the CodeSystemVersion using the value of dependantVersionEffectiveTime. Therefore we can remove the prevousDependencyRelease parameter. I will update the ticket for this. Thanks.

CoderMChu (changes requested):
It looks good but we can remove previousDependencyRelease parameter.

CoderMChu (approved):
It looks good to me. Thanks.

PR #397, opened 2022-05-03
ECL 2.0

kaicode:
Snowstorm implementation of ECL 1.4, 1.5, 1.6 and 2.0.

_I also made the FHIR unit tests more robust by moving the common data setup to a separate Spring Context, because those tests were getting in the way of my work._


pgwilliams:
I'll run this through our "feature of the week" Jenkins unit test job...

CoderMChu (approved):
@kaicode It looks great. Thanks for well documented and tested ECL features that cover versions from 1.4 to 2.0.

PR #403, opened 2022-05-12
MAINT-1961 Fix #402 bulk update hangs if branch not found.

### LINKED ISSUE #402: Possible bug on /concepts/bulk endpoint
Hi,
Would like to report a possible bug on the endpoint /browser/{branch}/concepts/bulk.

When I query for a branch that doesn't exist Snowstorm accepts the requests and reports the job ID as 'running' and is perpetually stuck in this status. There's nothing printed to the logs that helps me find or monitor the job (or otherwise indicating that there may be something wrong).

Expected behavior is to find branch or throw error and reject the request.

kaicode:
Fixes bug #402


CoderMChu (approved):
@kaicode It looks good. Thanks

PR #407, opened 2022-05-17
MAINT-1644 - Added copyConcepts method and a unit test

astro-snail:


src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +589,135 @@ public ExpressionStringPojo getConceptNormalForm(
...
+	@JsonView(value = View.Component.class)
+	public Collection<ConceptMini> copyConcepts(
+			@ApiParam(value = "Destination branch where the concepts are copied to.")
+			@PathVariable String branch,
+			@ApiParam(value = "Source branch where the concepts are selected from.")
+			@RequestParam String sourceBranch,
+			@ApiParam(value = "ECL expression for selecting concepts to copy between the branches.")
+			@RequestParam String ecl,
+			@ApiParam(value = "Include dependant components. Defaults to false.")
+			@RequestParam(required = false, defaultValue = "false") boolean includeDependencies,
```
QuyenLy87, line 604: Hi @astro-snail , it looks like you have not implemented this flag. Is it intentional?

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +589,135 @@ public ExpressionStringPojo getConceptNormalForm(
...
+			@RequestHeader(value = "Accept-Language", defaultValue = Config.DEFAULT_ACCEPT_LANG_HEADER) String acceptLanguageHeader) throws ServiceException {
```
QuyenLy87, line 605: acceptLanguageHeader seems to be a redundant parameter.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
astro-snail: Hi @QuyenLy87 yes, as it is not required to be implemented initially, as far as I understand. Will need more clarification on what are the dependencies in this case.

astro-snail: @QuyenLy87 I did implement the `includeDependencies` flag now.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
astro-snail: @QuyenLy87 Removed this parameter

src/main/java/org/snomed/snowstorm/core/pojo/LanguageDialect.java
```diff
@@ -48,8 +48,8 @@ public static List<String> toLanguageCodes(List<LanguageDialect> languageDialect
 		List<String> languageCodes = new ArrayList<>();
 		if (languageDialects != null) {
 			for (LanguageDialect languageDialect : languageDialects) {
-				if (StringUtils.isEmpty(languageDialect.getLanguageCode())) {
-					languageCodes.add(languageDialect.languageCode);
+				if (!StringUtils.isEmpty(languageDialect.getLanguageCode())) {
+					languageCodes.add(languageDialect.getLanguageCode());
```
CoderMChu, line 52: Well spotted.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +589,143 @@ public ExpressionStringPojo getConceptNormalForm(
...
+			try {
+				SAxiomRepresentation axiomRepresentation = axiomConversionService.convertAxiomMemberToAxiomRepresentation(axiomMember);
+				if (axiomRepresentation != null) {
+					Set<Relationship> relationships = axiomRepresentation.getLeftHandSideNamedConcept() != null
+							? axiomRepresentation.getRightHandSideRelationships()
+							: axiomRepresentation.getLeftHandSideRelationships();
+
+					for (Relationship relationship : relationships) {
+						String targetConceptId = relationship.getDestinationId();
+						Concept targetConcept = conceptService.find(targetConceptId, sourceBranchPath);
```
CoderMChu, line 660: @astro-snail  Shouldn't the branch path be destination branch instead of source branch?

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +589,143 @@ public ExpressionStringPojo getConceptNormalForm(
...
+							.collect(Collectors.toList()));
+				});
+
+		Concept conceptCreated = conceptService.create(conceptVersion, destinationBranchPath);
+
+		// Create reference set members
+		if (!referenceSetMembers.isEmpty()) {
+			referenceSetMemberService.createMembers(
+					destinationBranchPath,
+					referenceSetMembers.stream()
```
CoderMChu, line 713: I think it makes the code clearer to introduce a variable rather than using stream.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
astro-snail: @CoderMChu It may not exist on the destination branch, so I search on the source branch and then proceed according to the includeDependencies flag and moduleId.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +589,143 @@ public ExpressionStringPojo getConceptNormalForm(
...
+	}
+
+	private String getModuleId(String branchPath) {
+		Branch branch = branchService.findBranchOrThrow(branchPath, true);
+		final boolean isExtension = (branch.getMetadata() != null && !Strings.isEmpty(branch.getMetadata().getString(BranchMetadataKeys.DEPENDENCY_PACKAGE)));
+
+		return isExtension ? branch.getMetadata().getString(BranchMetadataKeys.DEFAULT_MODULE_ID) : Concepts.CORE_MODULE;
+	}
+
+	private ConceptMini copyConcept(ConceptMini concept, String sourceBranchPath, String destinationBranchPath, List<LanguageDialect> languageDialects, boolean includeDependencies) throws ServiceException {
```
CoderMChu, line 634: The method name is copyConcept but the actual implementation is more than just copy a concept from one branch to another. Maybe donateConcept or transferConcept might be more suitable.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +589,143 @@ public ExpressionStringPojo getConceptNormalForm(
...
+		if (!referenceSetMembers.isEmpty()) {
+			referenceSetMemberService.createMembers(
+					destinationBranchPath,
+					referenceSetMembers.stream()
+							.map(referenceSetMember -> {
+								ReferenceSetMember referenceSetMemberVersion = new ReferenceSetMember(
+										destinationModuleId,
+										referenceSetMember.getRefsetId(),
+										referenceSetMember.getReferencedComponentId());
+
```
CoderMChu, line 719: @astro-snail  I think we might need to keep the axiom refset member id during concept promotion?

CoderMChu (changes requested):
@astro-snail  The tests look to be testing the right thing. I have left some comments and let me know if you would like catch up tomorrow and have a quick discussion. Thanks.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
kaicode: There is tones of business logic here. This method should be moved to a service rather than living in a controller.

astro-snail: @kaicode I tried that, but I got an issue with context not found :-( I think it did not like me adding referenceSetMemberService to the ConceptService class, but I am not sure.

astro-snail: @kaicode I think that @Lazy did the trick. It appears to be working now. Has moved `donateConcept()` (aka `copyConcept()`) and `getModuleId()` into `ConceptService` class.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -805,4 +809,110 @@ public void addClauses(Set<String> conceptIds, Boolean active, BoolQueryBuilder
...
+							? axiomRepresentation.getRightHandSideRelationships()
+							: axiomRepresentation.getLeftHandSideRelationships();
+
+					for (Relationship relationship : relationships) {
+						String targetConceptId = relationship.getDestinationId();
+						Concept targetConcept = find(targetConceptId, sourceBranchPath);
+						if (targetConcept == null) {
+							throw new ServiceException("Axiom " + axiomMember.getId() + ": target concept " + targetConceptId + " does not exist on path " + sourceBranchPath);
+						}
+						if (includeDependencies && targetConcept.getModuleId().equals(sourceModuleId)) {
```
QuyenLy87, line 850: Hi @astro-snail , Do you check whether or not the target concept is released?

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
kaicode: @astro-snail sorry I missed these comments. I'm glad Lazy works. A separate "concept donation service" is an alternative to get around the wiring issue but sounds like you don't need it.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +580,40 @@ public ExpressionStringPojo getConceptNormalForm(
...
+	public Collection<ConceptMini> donateConcepts(
+			@ApiParam(value = "Destination branch where the concepts are copied to.")
+			@PathVariable String branch,
+			@ApiParam(value = "Source branch where the concepts are selected from.")
+			@RequestParam String sourceBranch,
+			@ApiParam(value = "ECL expression for selecting concepts to copy between the branches.")
+			@RequestParam String ecl,
+			@ApiParam(value = "Include dependant components. Defaults to false.")
+			@RequestParam(required = false, defaultValue = "false") boolean includeDependencies) throws ServiceException {
+		String sourceBranchPath = BranchPathUriUtil.decodePath(sourceBranch);
```
CoderMChu, line 596: @astro-snail It would be good if we add checks to make sure the source and destination are branches from different modules to prevent this being misused.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -580,6 +580,40 @@ public ExpressionStringPojo getConceptNormalForm(
...
+		QueryService.ConceptQueryBuilder queryBuilder = queryService.createQueryBuilder(true)
+				.activeFilter(true)
+				.ecl(ecl)
+				.isReleased(true)
+				.resultLanguageDialects(languageDialects);
+
+		List<ConceptMini> conceptsFound = queryService.search(queryBuilder, sourceBranchPath, LARGE_PAGE).getContent();
+		List<ConceptMini> conceptsDonated = new ArrayList<>();
+
+		for (ConceptMini concept : conceptsFound) {
```
CoderMChu, line 611: @astro-snail The ECL results potentially will return concepts are already in the core module. We should add logic here to prevent these being selected again.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
```diff
@@ -805,4 +809,110 @@ public void addClauses(Set<String> conceptIds, Boolean active, BoolQueryBuilder
...
+							donateConcept(new ConceptMini(targetConcept, languageDialects), sourceBranchPath, destinationBranchPath, languageDialects, true);
```
CoderMChu, line 851: @astro-snail This works for simple dependencies but might fail when there are circular dependencies. Let me check whether these complex cases do exist or not.

CoderMChu (commented):
@astro-snail Moving the logic to the service class looks good. I have added some minor comments. Thanks.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptService.java
CoderMChu: @astro-snail  I have checked and it is not possible to have circular dependencies. Please ignore this. Thanks.

PR #408, opened 2022-05-20
MAINT-1968 Fix code quality warnings and recommendation from LGTM report.

kaicode:
@CoderMChu I checked the LGTM report and found a couple of issues.
This PR fixes the 2 warnings and 1 recommendation but I'm not sure how best to tackle the 1 error that was raised. Let's talk further.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
@@ -227,15 +227,16 @@ public class QueryConcept extends DomainEntity<QueryConcept> {
 				|| !this.getAncestors().equals(other.getAncestors())) {
 			return false;
 		}
-		final Map<Integer, Map<String, List<Object>>> groupedAttributesMap = this.getGroupedAttributesMap();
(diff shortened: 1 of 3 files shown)


CoderMChu (approved):

CoderMChu:
@kaicode  We don't seem to use the values in "internalIdsToHide" after being collected. We could remove it to avoid the error about "The contents of this container are never accessed." Maybe this was added when we were debugging locally.

CoderMChu:
@kaicode After checking the commit history, we used to update the branch with this. However the internal id is added via the commit object now. https://github.com/IHTSDO/snowstorm/blob/develop/src/main/java/org/snomed/snowstorm/core/data/services/BranchMergeService.java#L494

CoderMChu:
@kaicode Thanks. I will fix the remaining error.

kaicode:
Thanks @CoderMChu !

kaicode:
@CoderMChu you are right that the internalId is being added to the commit, maybe they are only added to the `internalIdsToHide` collection for debugging or logging purposes. That doesn't look like it's needed 👍

PR #417, opened 2022-06-23
ISTO-42 Fix ECL caching bug

kaicode:
In the UAT environment these ECL queries are giving the same results.
<< 195967001 |Asthma| {{ D active = 0 }}
<< 195967001 |Asthma| {{ D active = 1 }}

> After investigation I found that this is caused by an ECL cache bug.
ECL is converted from Java objects back into text for the ECL cache lookup. This issue was that some of the description filters were not being included in the text when converted.

> I have now added [unit tests covering the conversion of all filter types](https://github.com/IHTSDO/snowstorm/blob/6d9bd593352277d80f5e3e1f373ecf601835e4c0/src/test/java/org/snomed/snowstorm/ecl/deserializer/ECLModelDeserializerServiceTest.java#L198) for all component types and fixed all remaining issues in this area. I found that the description module, effectiveTime and active filters were affected, meaning that changing any of these would have no effect on the ECL results.

All issues have been resolved and unit tested.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
 		<dependency>
 			<groupId>org.snomed.languages</groupId>
 			<artifactId>snomed-ecl-parser</artifactId>
-			<version>2.0.0</version>
+			<version>2.0.1-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>org.snomed.languages</groupId>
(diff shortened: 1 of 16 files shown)


kaicode:
This bug affects Snowstorm version 7.9.2 (with ECL 2.0), which at this point in time is in the `release-candidate` branch.

PR #418, opened 2022-06-23
PIP-182 Improve Snowstorm start up communication with AAG

markrynbeek:


pgwilliams (commented):
This change is going to prevent all commit listeners from firing before the ApplicationReadyEvent has fired isn't it?   There are a lot of them, maybe 10?   I'm not sure what all of their functions are.   It would seem safer to _just_ modify the AAG listener to be "pass through" when the application is not yet fully ready.

markrynbeek:
Apologies, I missed your comment. Not sure if I get it right, but I thought the class CommitServiceHookClient was configured specifically for AAG (service-hook.commit.url property) ?

PR #422, opened 2022-06-28
MAINT-1743 - Fixed loss of effective time after reverting changes in markDeletionsAndUpdates(); added/updated test cases

astro-snail:


pgwilliams (commented):
Wow, looks good.  Great to get that Description specific check out of there.

Just a comment about your check in comment, "changed" doesn't say very much.   Could you maybe say what the fix was eg "Fix loss of X in markDeletionsAndUpdates. Also added test cases"   

Similarly in your code comments, you've deleted  " Trying Concept module in attempt to restore effective time for the case where content has changed and then been reverted." which was expressive about what was going on.   You've added "Update effective time" and "Try to restore effective time", but that's sort of evident from the code.   What I want to know is:  _why_ are you trying to restore the effective time?  Why change the module id to be the same as that of the concept if you've only just set it to be the defaultModuleId?    Like, I know, and you know, but I don't think anyone who's not been staring at this for the last year would get it :-)

astro-snail:
> @pgwilliams thanks for the comment. Very good points. 

Initially the ticket was about fixing the effective time being cleared when we 
1) change the description in the extension module, 
2) save it, and then 
3) revert the change. 

The issue was that effective time was not being restored.

Looking at the code I saw that we try to change the module first to be the same as the concept and then (if effective time is still null) to the default module.

After a few discussions it was suggested to remove the method that was trying to define if we need to attempt this module changes, to fix the test, and to update effective time each time we attempt to change the module.

To be completely honest, I am still not quite sure why we are doing this module change when a component has changed.

Can I update the comment of the past commit?

astro-snail:
> Wow, looks good. Great to get that Description specific check out of there.
> 
> Just a comment about your check in comment, "changed" doesn't say very much. Could you maybe say what the fix was eg "Fix loss of X in markDeletionsAndUpdates. Also added test cases"
> 
> Similarly in your code comments, you've deleted " Trying Concept module in attempt to restore effective time for the case where content has changed and then been reverted." which was expressive about what was going on. You've added "Update effective time" and "Try to restore effective time", but that's sort of evident from the code. What I want to know is: _why_ are you trying to restore the effective time? Why change the module id to be the same as that of the concept if you've only just set it to be the defaultModuleId? Like, I know, and you know, but I don't think anyone who's not been staring at this for the last year would get it :-)

I will restore the expressive comment, it was a mistake to delete it, thanks for pointing that out!

pgwilliams:
By "Can I update the comment of the past commit?"   is that like "may I do this", or "how do I do this?"    The answer to the first is yes absolutely but make sure you're fully pulled before you force the push.... in your own branch this is less of a concern.  We usually warn each other when we do a force push so that people rebase immediately.    As to "How" you do it, I usually do rebase -i  HEAD~N    where N is the number of commits I might need to play with.   Then you get a text file to edit and you set an "r" against the commit you want to reword.   I can walk you through it if that was the question, rather than the first.

astro-snail:
> By "Can I update the comment of the past commit?" is that like "may I do this", or "how do I do this?" The answer to the first is yes absolutely but make sure you're fully pulled before you force the push.... in your own branch this is less of a concern. We usually warn each other when we do a force push so that people rebase immediately. As to "How" you do it, I usually do rebase -i HEAD~N where N is the number of commits I might need to play with. Then you get a text file to edit and you set an "r" against the commit you want to reword. I can walk you through it if that was the question, rather than the first.

@pgwilliams thank you! I have now updated this PR.

PR #423, opened 2022-06-29
MAINT-1910: add failing unit test for module jump and MAINT-1999 fix for same

markrynbeek:


src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1699,6 +1703,138 @@ private boolean waitUntil(Supplier<Boolean> supplier, int maxSecondsToWait) {
...
+			Description d = descriptionIterator.next();
+			memberService.createMember(mainBranchPath,
+				new ReferenceSetMember(d.getModuleId(), refSetId, d.getDescriptionId()));
+		}
+
+		// Create version
+		CodeSystem codeSystem = codeSystemService.find("SNOMEDCT");
+		codeSystemService.createVersion(codeSystem, 20210131, "20210131");
+
+		// Make a descriptor inactive
```
pgwilliams, line 1755: Don't confuse description with descriptor.  A _descriptor_ usually is in the context of a refset descriptor which is a refset that gives type information about what each column in a refset file relates to.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1699,6 +1703,138 @@ private boolean waitUntil(Supplier<Boolean> supplier, int maxSecondsToWait) {
...
+				.setLang("en").setLanguageCode("en").addLanguageRefsetMember(GB_EN_LANG_REFSET, Concepts.PREFERRED, true);
+
+		Concept concept = new Concept()
+				.addDescription(fsnDesc)
+				.addDescription(synonymDesc)
+				.addAxiom(new Relationship(ISA, parentConceptId));
+		concept = conceptService.create(concept, mainBranchPath);
+
+		// Add refset membership
+		Iterator<Description> descriptionIterator =	concept.getDescriptions().iterator();
```
pgwilliams, line 1744: This is a very generic way to add a refset member to a description...I'm not sure what it would mean.   Is it possible to go via the route of setting the acceptability for some specific language refset and save that way?

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1699,6 +1703,138 @@ private boolean waitUntil(Supplier<Boolean> supplier, int maxSecondsToWait) {
...
+
+		// Create version
+		CodeSystem codeSystem = codeSystemService.find("SNOMEDCT");
+		codeSystemService.createVersion(codeSystem, 20210131, "20210131");
+
+		// Make a descriptor inactive
+		Description desc = concept.getDescriptions().stream().filter(d -> d.getTerm().equals(synonymTerm)).findFirst().get();
+		desc.setActive(false);
+		desc.setInactivationIndicator(Concepts.inactivationIndicatorNames.get(DUPLICATE));
+		desc.setAssociationTargets(Maps.newHashMap("POSSIBLY_EQUIVALENT_TO", Sets.newHashSet(anotherDescriptionId)));
```
pgwilliams, line 1759: Although it doesn't make a difference for the purpose of this test, it's worth knowing that "Possibly Equivalent To" is an association used with concepts, not descriptions.    You should use "Refers To" instead.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1699,6 +1703,138 @@ private boolean waitUntil(Supplier<Boolean> supplier, int maxSecondsToWait) {
...
+		String anotherDescriptionId = "218680004";
+		String extensionModule = "45991000052106";
+		String refSetId="123459876000000000";
+
+		Map<String, String> intPreferred = Map.of(US_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED),
+				GB_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED));
+		Map<String, String> svPreferred = Map.of(swedishLanguageReferenceSet, descriptionAcceptabilityNames.get(PREFERRED));
+		Map<String, String> svAcceptable = Map.of(swedishLanguageReferenceSet, descriptionAcceptabilityNames.get(ACCEPTABLE));
+		String ci = "CASE_INSENSITIVE";
+		conceptService.batchCreate(Lists.newArrayList(new Concept(parentConceptId), new Concept(anotherConceptId)), mainBranchPath);
```
pgwilliams, line 1729: Does this just create a single IS_A ?

Add another inferred relationship as well please, so that we can inactivate that as well.    Some attribute - causative agent or something like that.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1699,6 +1703,138 @@ private boolean waitUntil(Supplier<Boolean> supplier, int maxSecondsToWait) {
...
+
+		// Create next version
+		codeSystemService.createVersion(codeSystem, 20210531, "20210531");
+
+		// Create an extension code system (Sweden)
+		codeSystemService.createCodeSystem(new CodeSystem("SNOMEDCT-SE", extensionBranchPath));
+		branchService.updateMetadata(extensionBranchPath, ImmutableMap.of(Config.DEFAULT_MODULE_ID_KEY, extensionModule, Config.DEFAULT_NAMESPACE_KEY, "1000052"));
+		concept = conceptService.find(concept.getConceptId(), extensionBranchPath);
+
+		assertFalse(concept.isActive());
```
pgwilliams, line 1784: Now please _reactivate_ the concept before adding the Swedish description. 

The test here is to see what happens to the _inactive_ inactivation indicators and historical associations.   So we need an active concept for that to happen (and then version again)

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1699,6 +1703,138 @@ private boolean waitUntil(Supplier<Boolean> supplier, int maxSecondsToWait) {
...
+				for (ReferenceSetMember rsm : members) {
+					if (extensionModule.equals(rsm.getModuleId())) {
+						countIntlDescrRefSetInExtensionModule++;
+					}
+				}
+			}
+		}
+		assertEquals(0, countIntlDescrRefSetInExtensionModule);
+
+		// Re-activate the concept
```
pgwilliams, line 1812: Well having an extension re-activate a core concept is an interesting case, and one we should probably look at.   But for this test, if you just move this code up to before you start working in the extension branch, then we have a more likely scenario.

If you reactivate in the extension, then ALL the components (except the active descriptions) will jump - quite correctly - into the extension module.    What we're trying to spot here is components doing that when they _shouldn't_.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -1699,6 +1703,138 @@ private boolean waitUntil(Supplier<Boolean> supplier, int maxSecondsToWait) {
...
+		concept.setActive(true);
+		transferredConcept = simulateRestTransfer(concept);
+		conceptService.update(transferredConcept, extensionBranchPath);
+		concept = conceptService.find(concept.getConceptId(), extensionBranchPath);
+
+		// Check the reference sets again
+		// Count: in reference member set, find the module id for the (international) FSN; this should not be the extension module
+		// but (bug) there is one anyway..
+		countIntlDescrRefSetInExtensionModule =0;
```
pgwilliams, line 1821: If you re-activate prior to adding the description, then you'll only need to check the modules once.

pgwilliams (commented):
That's a great unit test, Mark.

I've made several comments for tweaks.   I'm afraid I'm looking for something very specific with this data setup.    As you know, we don't see these module jumping errors _that_ often, given the amount of work that's being done.  So it's very much an edge case that we're trying to recreate.   Thanks!

PR #424, opened 2022-07-04
MAINT-1921 (Mitigate security risk - by upgrading Spring Boot library)

CoderMChu:


kaicode:
This PR contains Spring-Boot 2.6.6. Since this work was done Spring Boot 2.6.8 has become available. Based on the version numbers I expect that contains fixes. Version 2.7.0 is also now available.

CoderMChu:
@kaicode That is a very good point. I will update it to use 2.7.0. Thanks.

PR #429, opened 2022-07-20
Remove old redundant ECL parser generation files.

kaicode:
These files were from a long time ago, probably before the ecl-parser library was introduced. They are outdated (probably ECL 1.3) and not referenced anywhere in the code. Let's just delete them.


CoderMChu:
@kaicode  Thanks.

PR #430, opened 2022-07-29
FRI-443 Batch processing N concept changes per JMS message

markrynbeek:
There's also a related change made in traceability-service


src/main/java/org/snomed/snowstorm/core/data/services/traceability/TraceabilityConsumer.java
```diff
@@ -1,23 +1,56 @@
...
+	/**
+	 * A large activity object (with too many changes) should be messaged in batches.
+	 * @param activity
+	 */
+	void sendInBatches(Activity activity) {
+		int changeListSize = activity.getChanges().size();
+		logger.info("Number of concept activities is {}. larger than max ({}).", changeListSize, maxConceptActiviesPerMessage );
+		int idx = 0;
+
+		while (idx <= activity.getChanges().size() -1 ) {
```
CoderMChu, line 46: @markrynbeek  The logic is sound however we could make it simpler by using com.google.common.collect.Iterables.partition() method. The alternative batching is based on changes for N concepts rather than N changes.

src/main/resources/application.properties
```diff
@@ -454,6 +454,9 @@ spring.activemq.user=
 # Login password of the broker.
 spring.activemq.password=
 
+# Cap the amount of activities sent per message to prevent memory issues
+activemq.max.message.concept-activities=200000
```
CoderMChu, line 458: @markrynbeek  We should add this new config in the Authoring Traceability section with matching prefix.

src/main/java/org/snomed/snowstorm/core/data/services/traceability/TraceabilityConsumer.java
```diff
@@ -47,7 +47,7 @@ void sendInBatches(Activity activity) {
 			logger.info("Sending batch with start index {}.", idx);
 			List<Activity.ConceptActivity> changes = activity.getChanges().subList(idx, Math.min(idx + maxConceptActiviesPerMessage, changeListSize));
 			Activity activitySlice = new Activity(activity.getUserId(), activity.getBranchPath(),
-			activity.getCommitTimestamp(), activity.getSourceBranch(), activity.getActivityType());
+			activity.getCommitTimestamp(), activity.getSourceBranch(), activity.getActivityType(), false);
```
CoderMChu, line 50: @markrynbeek Shouldn't we send an activity with changeListComplete set to true in the last message?

CoderMChu (commented):
@markrynbeek The fix will solve the JMS message size issue but it could lead further issues in the traceability service. I will continue reviewing your changes in the traceability service. After that we can have a discussion about the alternatives. Thanks.

src/main/java/org/snomed/snowstorm/core/data/services/traceability/TraceabilityConsumer.java
CoderMChu: Please ignore this one.

src/main/java/org/snomed/snowstorm/core/data/services/traceability/TraceabilityConsumer.java
markrynbeek: I'll use that partition method, thanks. As for N: each change list element is a concept plus its component changes, so N changes and N concepts amounts to the same.

markrynbeek:
This branch is giving me peculiar git conflicts with 'develop', will make a new branch / PR

Issue #366, opened 2022-01-27
Unable to run Snowstorm 7.6.0


hurtigcodes:
Hi, 

I'm unable to run the new snowstorm-7.6.0.jar. Could this issue have any relation to recent #340? We're not using a classification service in the server.

I've tried both the jar distributed here as well as my own jar built from source code using maven.

Logs:
```
2022-01-27 11:39:07.141  WARN 7644 --- [main] ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'remoteClassificationServiceClient': Unsatisfied dependency expressed through field 'classificationStatusService'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.snomed.snowstorm.core.data.services.classification.ClassificationStatusService' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
2022-01-27 11:39:07.145  INFO 7644 --- [IdentifierCacheManagerDaemon] o.s.s.c.d.s.i.IdentifierCacheManager     : Identifier cache manager sleep interrupted.
2022-01-27 11:39:07.145  INFO 7644 --- [IdentifierCacheManagerDaemon] o.s.s.c.d.s.i.IdentifierCacheManager     : Identifier cache manager polling stopped.
2022-01-27 11:39:07.154  INFO 7644 --- [main] o.apache.catalina.core.StandardService   : Stopping service [Tomcat]
2022-01-27 11:39:07.154 DEBUG 7644 --- [java-sdk-http-connection-reaper] com.amazonaws.http.IdleConnectionReaper  : Reaper thread: 

java.lang.InterruptedException: sleep interrupted
	at java.base/java.lang.Thread.sleep(Native Method)
	at com.amazonaws.http.IdleConnectionReaper.run(IdleConnectionReaper.java:181)

2022-01-27 11:39:07.155 DEBUG 7644 --- [java-sdk-http-connection-reaper] com.amazonaws.http.IdleConnectionReaper  : Shutting down reaper thread.
2022-01-27 11:39:07.177  INFO 7644 --- [main] ConditionEvaluationReportLoggingListener : 

Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
[...]

PR #431, opened 2022-08-02
spring.config.additional-location

danka74:
Replace --spring.config.location with non-replacing --spring.config.additional-location.
https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.core.spring.config.additional-location
Addresses #366

### DIFF
--- a/docs/configuration-guide.md
+++ b/docs/configuration-guide.md
@@ -24,7 +24,7 @@ snowstorm.rest-api.readonly=true
 ```
 Then start Snowstorm with an extra JVM parameter:
 ```bash
---spring.config.location=application-local.properties
+--spring.config.additional-location=application-local.properties
 ```
 
 ### Using command line arguments


kaicode:
We are dropping the use of `spring.config.location` in favour of using the root context (`/`) and allowing Nginx to specify the context using `proxy_set_header X-Forwarded-Prefix /snowstorm/snomed-ct;`
That is already in the development environment and should come into effect in the next Snowstorm release.
Will that make this PR redundant @danka74 ?

danka74:
Sorry, X-Forwarded-Path is for the server.servlet.contextPath property, isn't it? There would still be configuration going on?

danka74:
So, I guess the PR is not redundant, with the contextPart as the exception. It's about the location of the configuration .properties file and that local properties should complement and not replace default. Or did I misunderstand?

kaicode:
Yes, you are right. Sorry for my first comment, that was misleading.

kaicode (approved):
A valuable documentation change. Looks good.

kaicode:
Thanks @danka74 !

PR #435, opened 2022-08-22
FRI-443 Batch processing N concept changes per JMS message

markrynbeek:
I opened a new PR because the other branch had some persistent git conflicts.

The idea is to send the activities in small batches (size set through a property) and store them as such in the traceability backend (no change required there, no re-assembling)

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -454,6 +454,9 @@ spring.activemq.user=
 # Login password of the broker.
 spring.activemq.password=
 
+# Cap the amount of activities sent per message to prevent memory issues
+activemq.max.message.concept-activities=250
+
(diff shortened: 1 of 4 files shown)


src/main/resources/application.properties
```diff
@@ -454,6 +454,9 @@ spring.activemq.user=
 # Login password of the broker.
 spring.activemq.password=
 
+# Cap the amount of activities sent per message to prevent memory issues
+activemq.max.message.concept-activities=250
```
CoderMChu, line 458: @markrynbeek  It would be better to change this config to authoring.traceability.message.max-concepts or something similar to this.

CoderMChu (approved):
@markrynbeek It looks good. Thanks.

PR #442, opened 2022-09-09
MAINT-1862 - Added searchAfter paging to SDottedExpressionConstraint

astro-snail:


CoderMChu (commented):
@astro-snail The fix looks good to me. It would be great if you can add a test case. Thanks.

astro-snail:
@CoderMChu Added a test case for pagination with searchAfter

CoderMChu (commented):
@astro-snail  You could either add searchAfter page request assertions in existing test method or extract data set up to a private method to avoid copy and paste. Apart from that it looks good. Thanks.

astro-snail:
> 

Will do that. Thanks. Just was not sure if 2 methods using the same code already justify a private setup method.

CoderMChu (approved):
@astro-snail Great it looks good. Thanks.

PR #449, opened 2022-10-28
MAINT-1940 - Prevent inactive historical associations switching module

dmcgihtsdo:
The purpose of MAINT-1940 is to prevent inactive historical associations switching module when rebasing with a merge conflict.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
@@ -499,7 +499,9 @@ public class ConceptUpdateHelper extends ComponentService {
 							.noneMatch(id -> id.equals(member.getMemberId()))) {
 
 				member.setActive(false);
-				member.markChanged();
+				if (!member.buildReleaseHash().equals(member.getReleaseHash())) {
+					member.markChanged();
+				}
 				
 				//Any change to a component in an extension needs to be done in the default module
 				if (member.isChanged() && (defaultModuleId != null && 
--- a/src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
@@ -33,7 +33,9 @@ import org.springframework.stereotype.Service;
 import org.springframework.util.StringUtils;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static org.elasticsearch.index.query.QueryBuilders.boolQuery;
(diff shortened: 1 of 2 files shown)


dmcgihtsdo:
@kaicode I have included you in this PR as we had an initial chat within an earlier commit (which is now gone as I have force pushed).

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -3630,6 +3630,405 @@ void assertReleaseProperties(
 		}
 	}
 
+	@Test
+	void testRebaseDoesNotResultInExtensionStealingAssociationMemberWhenNoConflict() throws ServiceException, InterruptedException {
```
dmcgihtsdo, line 3634: This test shows that when there is no conflict, inactive historical associations do not switch module.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -3630,6 +3630,405 @@ void assertReleaseProperties(
...
+		assertEquals(CORE_MODULE, associationMember.getModuleId());
+		assertVersioned(associationMember, 20220331);
+
+		associationMember = memberService.findMember(extProject, associationMemberId);
+		assertEquals(CORE_MODULE, associationMember.getModuleId());
+		assertVersioned(associationMember, 20220331);
+	}
+
+	@Test
+	void testRebaseDoesNotResultInExtensionStealingAssociationMemberWhenConflict() throws ServiceException, InterruptedException {
```
dmcgihtsdo, line 3765: This test shows the inactive historical association switches module when there is a conflict. To cause the conflict, the Extension changes the modelling via classification.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -3630,6 +3630,405 @@ void assertReleaseProperties(
...
+		assertEquals(CORE_MODULE, associationMember.getModuleId());
+		assertVersioned(associationMember, 20220331);
+
+		associationMember = memberService.findMember(extProject, associationMemberId);
+		assertEquals(CORE_MODULE, associationMember.getModuleId());
+		assertVersioned(associationMember, 20220331);
+	}
+
+	@Test
+	void testPromotionDoesNotResultInExtensionStealingAssociationMember() throws ServiceException, InterruptedException {
```
dmcgihtsdo, line 3904: This test shows the inactive historical association does not switch module when promoting.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -499,7 +499,9 @@ private <T> List<ReferenceSetMember> updateMetadataRefset(Map<String, Set<String
 							.noneMatch(id -> id.equals(member.getMemberId()))) {
 
 				member.setActive(false);
-				member.markChanged();
+				if (!member.buildReleaseHash().equals(member.getReleaseHash())) {
+					member.markChanged();
+				}
```
dmcgihtsdo, line 504: In an earlier conversation, @kaicode does not think this is the correct solution as we shouldn't care about the release hash.

CoderMChu: @dmcgihtsdo I agree with Kai's comment. I think we could do some refactoring in this method.

CoderMChu: @dmcgihtsdo  I will refactor this and we can catch up tomorrow.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -499,16 +491,12 @@ private <T> List<ReferenceSetMember> updateMetadataRefset(Map<String, Set<String
 							.noneMatch(id -> id.equals(member.getMemberId()))) {
 
 				member.setActive(false);
-				if (!member.buildReleaseHash().equals(member.getReleaseHash())) {
-					member.markChanged();
-				}
-				
-				//Any change to a component in an extension needs to be done in the default module
+				member.markChanged();
```
pgwilliams, line 494: Are we definitely sure that only changed members will appear through this code path?   What about if we've changed it back so that it now equals the previous release hash?

CoderMChu: I think the logic about whether a component has changed is fine even though this method requires refactoring as well. However what is the correct module id to use below from line 505 won't work when reverting extension changes for core module components.

PR #451, opened 2022-11-15
FRI-565 Correct changeType from DELETE to INACTIVATE when upgrading CodeSystem

dmcgihtsdo:


src/main/java/org/snomed/snowstorm/core/data/services/BranchMergeService.java
```diff
@@ -396,7 +396,7 @@ private <T extends SnomedComponent<T>> void removeRebaseDuplicateVersions(Class<
 			ElasticsearchRepository repository = domainEntityConfiguration.getComponentTypeRepositoryMap().get(componentClass);
 			logger.info("Taking parent version of {} {}s on {}", duplicateComponents.size(), componentClass.getSimpleName(), path);
 			versionControlHelper.endOldVersionsOnThisBranch(componentClass, duplicateComponents, idField, clause, commit, repository);
-			BranchMetadataHelper.getRebaseDuplicatesRemoved(commit).put(componentClass.getSimpleName(), duplicateComponents);
+			BranchMetadataHelper.setRebaseDuplicatesRemoved(commit, componentClass, duplicateComponents);
```
dmcgihtsdo, line 399: I replaced this line to use the new util method.

src/main/java/org/snomed/snowstorm/core/data/services/BranchMergeService.java
```diff
@@ -487,6 +487,7 @@ private void fixDuplicateComponentsOfType(String branch, Commit commit, BranchCr
 					// End duplicate components in extension module
 					extensionVersion.setEnd(commit.getTimepoint());
 					repository.save(extensionVersion);
+					BranchMetadataHelper.setRebaseDuplicatesRemoved(commit, extensionVersion.getClass(), Set.of(extensionVersion.getId()));
```
dmcgihtsdo, line 490: This will store the duplicates "in memory" of the commit, which will later be accessed when writing traceability logs.

src/test/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceTest.java
```diff
@@ -644,6 +651,86 @@ void testRebaseWithChangeSupersededByOtherProject() throws Exception {
...
+		// Version International
+		codeSystem = codeSystemService.find("SNOMEDCT");
+		codeSystemService.createVersion(codeSystem, 20220731, "20220731");
+
+		// Extension fixes modelling (through classification)
+		concept = conceptService.find(carId, extMain);
+		concept.getRelationships().iterator().next().setActive(false).updateEffectiveTime();
+		concept.getRelationships().add(new Relationship(ISA, vehicleId));
+		concept = conceptService.update(concept, extMain);
+		List<String> extRelationships = getRelationshipIds(concept);
```
CoderMChu, line 704: @dmcgihtsdo  Good test. But I think the traceability assertion for inactivation should be here before upgrade.

src/test/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceTest.java
```diff
@@ -644,6 +651,86 @@ void testRebaseWithChangeSupersededByOtherProject() throws Exception {
...
+		codeSystem = codeSystemService.find("SNOMEDCT-TEST");
+		codeSystemUpgradeService.upgrade(codeSystem, 20220731, false);
+
+		// Extension ends up with International version of inactivated Relationship post upgrade
+		relationship = relationshipService.findRelationship(extMain, relationshipToBeInactivated);
+		assertTrue(relationship.isReleased());
+		assertEquals(Concepts.CORE_MODULE, relationship.getModuleId());
+		assertEquals(20220731, relationship.getReleasedEffectiveTime());
+
+		// Assert
```
CoderMChu, line 716: After upgrade the traceability activity should be "superseded" for extension, because the version from international release is chosen.

src/main/java/org/snomed/snowstorm/core/data/services/BranchMetadataHelper.java
```diff
@@ -69,6 +69,14 @@ public static Map<String, Set<String>> getRebaseDuplicatesRemoved(Commit commit)
 		return (Map<String, Set<String>>) commit.getBranch().getMetadata().getAsMap().computeIfAbsent(key, k -> new HashMap<String, Set<String>>());
 	}
 
+	public static void setRebaseDuplicatesRemoved(Commit commit, Class<?> clazz, Set<String> duplicateComponentIds) {
+		if (commit == null || clazz == null || duplicateComponentIds == null || duplicateComponentIds.isEmpty()) {
```
CoderMChu, line 73: @dmcgihtsdo This is a bad idea. You could just add items to the map after getRebaseDuplicatesRemoved

src/main/java/org/snomed/snowstorm/core/data/services/BranchMergeService.java
CoderMChu: Well spot it.

CoderMChu (changes requested):
@dmcgihtsdo It looks good apart from minor changes required. Thanks.

src/test/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceTest.java
dmcgihtsdo: Mm, how do you mean @CoderMChu? 

Before the upgrade, the Extension will inactivate which goes through as INACTIVATE. During the upgrade, Snowstorm identifies there is a duplicate edit and discards the Extension's change. This eventually leads to the traceability message being DELETE instead of INACTIVATE.

So I need the assertion after the upgrade?

src/test/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceTest.java
dmcgihtsdo: Thanks @CoderMChu.

I have added this as an assertion.

src/main/java/org/snomed/snowstorm/core/data/services/BranchMetadataHelper.java
CoderMChu: @dmcgihtsdo It would be cleaner if you just add these items directly into the map. It is a bad idea to add setters in a helper class.

src/test/java/org/snomed/snowstorm/core/data/services/TraceabilityLogServiceTest.java
CoderMChu: @dmcgihtsdo This bug is about wrong change type used when extension inactivates international content. This can happen during classification for relationships. In your test case the inactivation happens before upgrade, that's why I added a comment. Your assumption about "During the upgrade, Snowstorm identifies there is a duplicate edit and discards the Extension's change. This eventually leads to the traceability message being DELETE instead of INACTIVATE." is not correct. Discarding changes during rebase will only lead to changes to be superseded. The change type of a traceability is determined by the nature of change for a given component. This can happen for activity type "CONTENT_CHANGE" or "CLASSIFICATION_SAVE". If you are still not clear we can have a quick call.

src/main/java/org/snomed/snowstorm/core/data/services/BranchMetadataHelper.java
dmcgihtsdo: Great, I have done this.

PR #459, opened 2022-11-23
Updates to editions and direct aliases.

JonZammit:
As requested by @kaicode, I have made config changes to support the Swiss Edition in Code Systems and Known Dialects.

Note that the Swiss edition appears to have 3 language refsets in use - French, Italian, and German.

Also made changes to Known Dialects to support the ECL dialect format change. (See addition of "ISO-x" codes.)


kaicode:
Hi @JonZammit, the ECL dialect change should not be added until Snowstorm implements ECL 2.1. Let's hold those changes back please.

kaicode:
I've added another commit to this PR to add this breaking change to the changelog. The Feb release will include many other breaking changes in the FHIR API rewrite so we will need to move to Snowstorm version 8 anyway. It's a good time for breaking changes.

PR #460, opened 2022-11-23
RT2-126, 130, and 259 for SNOMED develop

RichardAWood:
This is for adding RT2 functionality to the develop branch, whenever the team is ready to receive it.


src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -133,14 +137,16 @@ public ItemsPage<?> findConcepts(
 			@RequestParam(required = false) Boolean isNullEffectiveTime,
 			@RequestParam(required = false) Boolean isPublished,
 			@RequestParam(required = false) String statedEcl,
+			@RequestParam(required = false, defaultValue = "false") Boolean includeLeafFlag,
+			@RequestParam(defaultValue = "inferred") Relationship.CharacteristicType form,
```
kaicode, line 141: There is a conflict here between the parameters `ecl`, `statedEcl` and `form` (which can be inferred or stated).
To ensure consistency perhaps the `statedEcl` parameter should be deprecated and the stated form used consistently across parameters when an argument of `form` = stated is given.

src/main/java/org/snomed/snowstorm/rest/ConceptController.java
```diff
@@ -554,12 +587,50 @@ public Collection<ConceptMini> findConceptParents(@PathVariable String branch,
...
+			conceptMiniMap.get(conceptId).addExtraField("descriptions", conceptMiniMap.get(conceptId).getActiveDescriptions());
+			conceptMiniMap.get(conceptId).addExtraField("ancestorPath", ancestorPath);
+			conceptsWithAncestorPaths.add(conceptMiniMap.get(conceptId));
+		}
+		
+		return conceptsWithAncestorPaths;
+	}	
+	
+	private ArrayList<ConceptMini> ancestorPathHelper(String branch, Relationship.CharacteristicType form, String conceptId, ArrayList<ConceptMini> pathSoFar, String acceptLanguageHeader) throws ServiceException {
+		Collection<ConceptMini> conceptParents = findConceptParents(branch, conceptId, form, false, acceptLanguageHeader);
```
kaicode, line 621: **Performance Observation**
Currently the method iterates through concepts and for each concept hits Elasticsearch in a recursive method to load ancestor paths. 
For 10 concepts with an average depth in the hierarchy of 7 there will be at least 70 Elasticsearch calls. I assume this is very slow and gets progressively slower with more concepts.

**Optimisation Suggestion**
We could load the ancestor paths of around 500 concepts at once using just 2 Elasticsearch calls
- The first query can load the unordered set of ancestors for all concepts at once
  - Use a query against the QueryConcept type with `must(term("stated", form == stated)).must(terms("conceptIdL, allConcepts))`
  - Gather concept ids from `ancestors` field
- The second query can load the direct parents of all the relevant concepts
  - Query against QueryConcept type again with `must(term("stated", form == stated)).must(terms("conceptIdL, allConceptsPlusAncestors))`
  - From this a hierarchy can be created in memory. The ancestor paths can be extracted from this.

RichardAWood:
Closing this, as it will be replaced with a new PR once the content is validated on Integration.

PR #464, opened 2022-11-29
FRI-550 Implemented new asyn API for code system upgrade

QuyenLy87:


src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemUpgradeService.java
```diff
@@ -50,24 +57,89 @@ public class CodeSystemUpgradeService {
 	@Value("${snowstorm.rest-api.readonly}")
 	private boolean isReadOnly;
 
+	private static final Map<String, CodeSystemUpgradeJob> upgradeJobMap = new HashMap<>();
+
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
+	public CodeSystemUpgradeService() {
+		crunchifyTimer = new Timer();
```
pgwilliams, line 65: Why call it a crunchifyTimer here please @QuyenLy87 ?   Does that mean something?  It's just that the object created is just a Timer, so it's like you're saying something extra, but I don't know what crunchify means?

src/main/java/org/snomed/snowstorm/core/data/services/pojo/CodeSystemUpgradeJob.java
```diff
@@ -0,0 +1,53 @@
+package org.snomed.snowstorm.core.data.services.pojo;
+
+public class CodeSystemUpgradeJob {
+
+	public enum UpgradeStatus {
+		RUNNING, COMPLETED, FAILED;
```
pgwilliams, line 6: Nice, very clean.

src/main/java/org/snomed/snowstorm/rest/CodeSystemController.java
```diff
@@ -178,9 +181,21 @@ public CodeSystemVersion updateVersion(@PathVariable String shortName, @PathVari
 					"to support creating a new version of the extension. \n\n" +
 					"If you are the extension maintainer an integrity check should be run after this operation to find content that needs fixing. ")
 	@PostMapping(value = "/{shortName}/upgrade")
-	public void upgradeCodeSystem(@PathVariable String shortName, @RequestBody CodeSystemUpgradeRequest request) throws ServiceException {
+	public ResponseEntity<Void> upgradeCodeSystem(@PathVariable String shortName, @RequestBody CodeSystemUpgradeRequest request) throws ServiceException {
```
pgwilliams, line 184: Good use of ResponseEntity to deal with both sync and async styles of call.

pgwilliams: As discussed, I note that we're removing the option to use this method in a synchronous manner.   I think that does simplify the problem greatly.

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemUpgradeService.java
```diff
@@ -50,24 +57,89 @@ public class CodeSystemUpgradeService {
...
+	@PreAuthorize("hasPermission('ADMIN', #codeSystem.branchPath)")
+	@Async
+	public void upgradeAsync(String id, CodeSystem codeSystem, Integer newDependantVersion, boolean contentAutomations, SecurityContext securityContext) throws  ServiceException {
+		SecurityContextHolder.setContext(securityContext);
+		upgrade(id, codeSystem, newDependantVersion, contentAutomations);
+	}
+
 	@PreAuthorize("hasPermission('ADMIN', #codeSystem.branchPath)")
-	public void upgrade(CodeSystem codeSystem, Integer newDependantVersion, boolean contentAutomations) throws  ServiceException {
+	public void upgrade(String id, CodeSystem codeSystem, Integer newDependantVersion, boolean contentAutomations) throws  ServiceException {
```
pgwilliams, line 109: HashMap is not thread safe, but even if you were using ConcurrentHashMap, we've still got the problem of two calls coming in here at the same time even before they look at that map.    Now that you're working asynchronously, I think this method should return quickly enough that you don't need to worry about it blocking, so I think just a "synchronized" keyword in the definition of this function should protect us from race conditions.   Do you agree @CoderMChu ?

pgwilliams (changes requested):
Looks really nice @QuyenLy87.

The only thing I'm keen to see before we merge is the addition of they synchronized keyword on the service method, as mentioned in comments.  

Once that's in, I'm happy to merge.   Great work!

pgwilliams (approved):
Thank you for fully addressing my comments @QuyenLy87 , approved.

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemUpgradeService.java
CoderMChu: It is thread safe to add "synchronized" to the method. However we should try to avoid using "synchronized" keyword if possible at the method level. The synchronised block should be very small to avoid blocking. It depends on the use case I think. If we only allow one upgrade at a given time then adding "synchronized" will work. But if we want to allow upgrading two different CodeSystems concurrently then this will perform poorly.

PR #468, opened 2022-12-01
RT2-1505 - attempt to get multisearch ECL working again.

RichardAWood:
We discovered today that the ECL parameter we had added to MultiSearchService.findDescriptions() was no longer functioning after rebasing from develop, due to the multiple changes in the MultiSearchService and in the ECLQueryService.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
@@ -72,8 +72,9 @@ public class MultiSearchService implements CommitListener {
 
 		if (ecl != null) {
 			// ECL -> conceptIds
-			MultiBranchCriteria branchesQuery = getBranchesQuery();
-			Page<Long> page = eclQueryService.selectConceptIds(ecl, branchesQuery, false, LARGE_PAGE);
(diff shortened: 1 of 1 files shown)


pgwilliams:
Hi Rick

This hard coding of "MAIN" looks more like something you'd try out on your local machine to provide some diagnosis or temporary workaround, rather than something to check in to a branch.    Can we get a little detail on what the plan is here please?

RichardAWood:
You're absolutely correct - this is our attempt to get a 90% solution in to address a failure while we work on a true fix (which is also why we didn't submit a matching PR to the develop branch).

The issue that we're trying to resolve is that previously we had been taking the BoolQueryBuilder that getBranchesQuery() returned, and using it as the branchCriteria for a call to eclQueryService.selectConceptIds().  This worked until we rebased from develop.  Now getBranchesQuery() returns a MultiBranchCriteria instead, and when we send it to eclQueryService.selectConceptIds() we get an error:

<Map>
<error>BAD_REQUEST</error>
<message>Branch 'all-released' does not exist.</message>
</Map>

Any insight into the correct way to send multi-branch-criteria to the eclQueryService would be greatly appreciated - we couldn't find anything similar in the code to draw from, and switching to the BranchCriteria was the best we could come up with in a hurry.

pgwilliams:
Done!  Thanks for the extra detail. :-)

PR #480, opened 2023-02-08
FRI-491 Improve automerge function

dmcgihtsdo:
FRI-491 is concerned with improving the middle column of the conflict window.


src/main/java/org/snomed/snowstorm/core/data/domain/Axiom.java
```diff
@@ -119,12 +120,49 @@ public Axiom setRelationships(Set<Relationship> relationships) {
 		return this;
 	}
 
-	@JsonIgnore
 	public ReferenceSetMember getReferenceSetMember() {
 		return referenceSetMember;
 	}
```
dmcgihtsdo, line 125: I've removed this annotation because I needed the Axiom's ReferenceSetMember whilst constructing the middle column.

src/main/java/org/snomed/snowstorm/core/data/domain/Axiom.java
```diff
@@ -119,12 +120,49 @@ public Axiom setRelationships(Set<Relationship> relationships) {
...
 	public void setReferenceSetMember(ReferenceSetMember referenceSetMember) {
 		this.referenceSetMember = referenceSetMember;
 	}
+
+    public void clone(Axiom axiom) {
+        setAxiomId(axiom.getAxiomId());
+        setModuleId(axiom.getModuleId());
+        setActive(axiom.isActive());
+        setReleased(axiom.isReleased());
+        setDefinitionStatusId(axiom.getDefinitionStatusId());
```
dmcgihtsdo, line 136: I've added a clone method to all of the components. These clone methods are used so that the "auto-merged version" of the Concept is based on the LHS of the conflict window.

src/main/java/org/snomed/snowstorm/core/data/domain/Concepts.java
```diff
@@ -195,5 +195,9 @@ public class Concepts {
 	public static final String HEART_STRUCTURE = "80891009";
 	public static final String REFERENCED_COMPONENT = "449608002";
 	public static final String CONCEPT_TYPE_COMPONENT = "900000000000461009";
+	public static final String HAS_FILLING = "827081001";
+	public static final String PHYSICAL_OBJECT = "260787004";
+	public static final String DEVICE = "49062001";
+	public static final String DOMESTIC = "272179000";
```
dmcgihtsdo, line 201: These constants have been added for tests.

src/main/java/org/snomed/snowstorm/core/data/domain/Description.java
```diff
@@ -267,6 +267,17 @@ public Set<ReferenceSetMember> getLangRefsetMembers() {
...
+	public ReferenceSetMember getLangRefsetMember(String memberId) {
+		for (ReferenceSetMember referenceSetMember : langRefsetMembersMap.values().stream().flatMap(Collection::stream).collect(Collectors.toSet())) {
+			if (memberId.equals(referenceSetMember.getMemberId())) {
+				return referenceSetMember;
+			}
+		}
+
+		return null;
+	}
+
```
dmcgihtsdo, line 280: I added this method for some tests.

src/main/java/org/snomed/snowstorm/core/data/domain/ReferenceSetMember.java
```diff
@@ -191,16 +191,18 @@ public String getMemberId() {
 		return memberId;
 	}
 
-	public void setMemberId(String memberId) {
+	public ReferenceSetMember setMemberId(String memberId) {
 		this.memberId = memberId;
+		return this;
 	}
```
dmcgihtsdo, line 197: I changed the return type so I could chain the construction of the object.

src/main/java/org/snomed/snowstorm/core/data/domain/Axiom.java
```diff
@@ -119,12 +120,49 @@ public Axiom setRelationships(Set<Relationship> relationships) {
 		return this;
 	}
 
-	@JsonIgnore
```
kaicode, line 122: I am not comfortable exposing the axiom refset member in the REST API. We are currently hiding this, and the OWL expression contained within, in favour of viewing and editing concepts in terms of axioms with a set of generated "relationships".
Editing the OWL expression within the refset and sending it back to the API will not work as intended so I think it's better to hide the refset member. CC @CoderMChu

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
```diff
@@ -0,0 +1,518 @@
...
+     *
+     * @param sourceConcept    The state that will be the baseline for the new, auto-merged Concept.
+     * @param targetConceptOld The state to help identify which edits belong to targetConceptNew.
+     * @param targetConceptNew Edits belonging to this Concept will be re-applied to the new, auto-merged Concept.
+     * @return A new instance of Concept which is a combination of both sourceConcept and targetConceptNew.
+     */
+    public Concept autoMerge(Concept sourceConcept, Concept targetConceptOld, Concept targetConceptNew) {
+        if (sourceConcept == null || targetConceptOld == null || targetConceptNew == null) {
+            return new Concept();
+        }
```
dmcgihtsdo, line 29: I have replaced the existing "automerge" function with this class. Having the logic in its own class let me write more tests. All current tests pass with this new (moderately the same) implementation.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
```diff
@@ -0,0 +1,518 @@
...
+            }
+        }
+
+        return identifiers;
+    }
+
+    // Base auto-merged Concept on sourceConcept; edits on targetConceptNew will be re-applied.
+    private Concept rebaseTargetChangesOntoSource(Concept sourceConcept, Concept targetConceptOld, Concept targetConceptNew, Map<ComponentType, Set<String>> changesOnTargetConceptNew) {
+        Concept mergedConcept = new Concept();
+        mergedConcept.clone(sourceConcept); // mergedConcept is deeply based on sourceConcept
```
dmcgihtsdo, line 170: This is where the new clone methods are called.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
```diff
@@ -0,0 +1,518 @@
...
+        mergedConcept.setDefinitionStatus(getValueChanged(targetConceptOld.getDefinitionStatus(), targetConceptNew.getDefinitionStatus(), sourceConcept.getDefinitionStatus()));
+        mergedConcept.setReleaseHash(getValueChanged(targetConceptOld.getReleaseHash(), targetConceptNew.getReleaseHash(), sourceConcept.getReleaseHash()));
+        mergedConcept.setReleased(getValueChanged(targetConceptOld.isReleased(), targetConceptNew.isReleased(), sourceConcept.isReleased()));
+
+        mergedConcept.updateEffectiveTime();
+
+        // If changed, ensure component is re-activated unless the RHS had the inactivation.
+        if (mergedConcept.getEffectiveTime() == null && !sourceConcept.isActive() && targetConceptNew.isActive()) {
+            mergedConcept.setActive(true);
+        }
```
dmcgihtsdo, line 206: Without this block (and the others like it), the function basically takes the Source concept and what has changed on the Target concept and combines the two. As a result, we could end up with the middle column having a different module but also remaining inactive. This block re-activates the Concept (and other components in their respective methods) if there has been a change.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
```diff
@@ -0,0 +1,518 @@
...
+
+    // valueDefault is coming from Source and may have newer/different value than valueOld.
+    private Integer getValueChanged(Integer valueOld, Integer valueNew, Integer valueDefault) {
+        boolean valueHasChanged = !Objects.equals(valueOld, valueNew);
+        if (valueHasChanged) {
+            return valueNew;
+        } else {
+            return valueDefault;
+        }
+    }
```
dmcgihtsdo, line 517: These are util methods for working out what the author has changed. If the author hasn't changed the field, then use the field from the Source, i.e. valueDefault.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -243,58 +258,6 @@ private Set<Long> getConflictingConceptIds(MergeReview mergeReview) {
 		return Sets.intersection(sourceToTargetReviewChanges, targetToSourceReviewChanges);
 	}
 
-	private Concept autoMergeConcept(Concept sourceConcept, Concept targetConcept) {
-		final Concept mergedConcept = new Concept();
-
-		// In each component favour the source version unless only the target is unpublished
-		Concept winningConcept = sourceConcept.getEffectiveTimeI() != null && targetConcept.getEffectiveTimeI() == null ? targetConcept : sourceConcept;
```
dmcgihtsdo, line 250: This is the old method that has been replaced by the automerge class.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -4245,6 +4245,271 @@ void testPromotionDoesNotResultInExtensionStealingAssociationMember() throws Ser
...
+		assertEquals(CORE_MODULE, concept.getModuleId());
+		assertEquals(3, concept.getDescriptions().size());
+	}
+
+	@Test
+	void testAutoMergeWhenRebasingVersionAhead() throws ServiceException, InterruptedException {
+		String intMain = "MAIN";
+		String extMain = "MAIN/SNOMEDCT-TEST";
+		Map<String, String> intPreferred = Map.of(US_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED), GB_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED));
+		Map<String, String> intAcceptable = Map.of(US_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED), GB_EN_LANG_REFSET, descriptionAcceptabilityNames.get(ACCEPTABLE));
```
dmcgihtsdo, line 4350: I think this is the test that fails on develop. 

Essentially: 

- lhs changes the concept's module (moves from module A to module B)
- rhs inactivates the concept (currently in module A)
- middle column ends up inactive in module A (because this is the module the concept is in on the task)

This is wrong because we're suggesting inactivation and also moving the module back to module A. With the new class, the middle column ends up inactive in module B.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -4245,6 +4245,271 @@ void testPromotionDoesNotResultInExtensionStealingAssociationMember() throws Ser
...
+		assertEquals(20220228, concept.getReleasedEffectiveTime()); // Concept has moved along
+		assertNull(concept.getEffectiveTimeI()); // Changes have been applied to the Concept
+		assertEquals(CORE_MODULE, concept.getModuleId()); // This is what changed on LHS
+		assertFalse(concept.isActive()); // This is what changed on RHS
+	}
+
+	@Test
+	void testReappliedChangeShouldBeLegalChange() throws ServiceException, InterruptedException {
+		String intMain = "MAIN";
+		String extMain = "MAIN/SNOMEDCT-TEST";
```
dmcgihtsdo, line 4432: This test shows that if the lhs inactivates the concept and the rhs changes something, the middle column is reactivated.

src/main/java/org/snomed/snowstorm/core/data/domain/Axiom.java
dmcgihtsdo: Thanks for your input @kaicode. 

I needed this as the member was not being written to the store as part of the "ManuallyMergedConcept" string. As a result, the autoMerge code was falling over with a NPE because there was no ReferenceSetMember associated with the Axiom.

kaicode: It's not okay to expose this refset member in the REST API. If it is being exposed by the removal of this annotation I think we should look for an alternative approach. 
The refset member id is probably already persisted so you could load it from the branch when applying the merge?

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -169,7 +174,12 @@ public Collection<MergeReviewConceptVersions> getMergeReviewConflictingConcepts(
 				} else {
 					if (sourceVersion != null && targetVersion != null) {
 						// Neither deleted, auto-merge.
-						mergeVersion.setAutoMergedConcept(autoMergeConcept(sourceVersion, targetVersion));
+						Concept targetVersionBefore = conceptService.find(
+								sourceVersion.getConceptId(),
+								DEFAULT_LANGUAGE_DIALECTS,
+								ControllerHelper.parseBranchTimepoint(mergeReview.getTargetPath() + "@^")
+						);
```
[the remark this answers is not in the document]
CoderMChu: Can we hide the logic of targetVersionBefore inside autoMerge() method? e.g autoMerge(sourceVersion, targetVersion)

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
CoderMChu: Good catch.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
CoderMChu: Do we need to reactivate the concept here just for the module change? The project version(inactivation) is published already. We could just apply the module change in the middle column.

CoderMChu (commented):
@dmcgihtsdo I have added a couple of comments. We can catch up tomorrow if you have any questions. Thanks.

src/main/java/org/snomed/snowstorm/core/data/domain/Axiom.java
dmcgihtsdo: Thanks both, I'll look into a different way.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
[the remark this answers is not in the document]
dmcgihtsdo: Sure, I'll make that change @CoderMChu.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
dmcgihtsdo: Thanks @CoderMChu, if we don't re-activate though, aren't we then allowing the middle column to save invalid content? For example, the concept is inactive but moved module... validation will pick this up as a "no active state found in previous release" failure?

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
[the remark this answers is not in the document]
dmcgihtsdo: I have made that change @CoderMChu, although I had to supply the branch path as a parameter too.

src/main/java/org/snomed/snowstorm/core/data/domain/Concepts.java
dmcgihtsdo: I have made that change @CoderMChu.

src/main/java/org/snomed/snowstorm/core/data/domain/Axiom.java
dmcgihtsdo: @kaicode @CoderMChu I have implemented another way. I now find the member from elasticsearch rather than exposing the member in the rest api. You can see this in the `joinAxiomReferenceSetMembers` method in AutoMerger.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
```diff
@@ -0,0 +1,542 @@
...
+     * on which targetConcept has been edited must be supplied to help identify which fields of targetConcept have actually changed.
+     *
+     * @param sourceConcept The state that will be the baseline for the new, auto-merged Concept.
+     * @param targetConcept Edits belonging to this Concept will be re-applied to the new, auto-merged Concept.
+     * @param branchPath    The branch path on which targetConcept has changed.
+     * @return A new instance of Concept which is a combination of both sourceConcept and targetConcept.
+     */
+    public Concept autoMerge(Concept sourceConcept, Concept targetConcept, String branchPath) {
+        // Data required for auto-merging.
+        joinAxiomReferenceSetMembers(targetConcept, branchPath);
```
CoderMChu, line 38: @dmcgihtsdo What about GCIs?

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -4245,6 +4245,338 @@ void testPromotionDoesNotResultInExtensionStealingAssociationMember() throws Ser
...
+		concept = conceptService.find(vehicleId, taskB);
+		assertEquals(20220228, concept.getReleasedEffectiveTime()); // Concept has moved along
+		assertNull(concept.getEffectiveTimeI()); // Changes have been applied to the Concept
+		assertEquals(CORE_MODULE, concept.getModuleId()); // This is what changed on LHS
+		assertTrue(concept.isActive()); // Concept change should result in re-activation
+	}
+
+	@Test
+	void testRebasingVersionedAxiomChange() throws ServiceException, InterruptedException {
+		String intMain = "MAIN";
```
dmcgihtsdo, line 4515: @CoderMChu this is another test that fails on develop but passes on this feature branch.

dmcgihtsdo: Develop ends up with 2 axioms and lost release statuses.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
dmcgihtsdo: @CoderMChu , GCIs are included in the `getComponentsChangedOnTargetNew` method.

CoderMChu: Don't they need to be loaded from refset members the same way as class axioms?

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
```diff
@@ -0,0 +1,542 @@
...
+
+        mergedConcept.updateEffectiveTime();
+
+        // If changed, ensure component is re-activated unless the RHS had the inactivation.
+        if (mergedConcept.getEffectiveTime() == null && !sourceConcept.isActive() && targetConceptNew.isActive()) {
+            mergedConcept.setActive(true);
+        }
+    }
+
+    private void reapplyDescriptionChanges(Map<String, SnomedComponent<?>> sourceDescriptions, Map<String, SnomedComponent<?>> targetDescriptionsOld, Map<String, SnomedComponent<?>> targetDescriptionsNew, Set<String> changedDescriptionIds, Concept mergedConcept) {
```
CoderMChu, line 233: SnomedComponent<?> should be Description type.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
```diff
@@ -0,0 +1,542 @@
...
+            Axiom axiom = entrySet.getValue();
+
+            // There are no target edits if the identifier is absent; therefore, take from source.
+            mergedAxioms.putIfAbsent(unchangedAxiomId, axiom);
+        }
+
+        return new HashSet<>(mergedAxioms.values());
+    }
+
+    private void reapplyLangMemberChanges(Map<String, SnomedComponent<?>> sourceLangMembers, Map<String, SnomedComponent<?>> targetLangMembersOld, Map<String, SnomedComponent<?>> targetLangMembersNew, Set<String> changedLangMemberIds, Concept mergedConcept) {
```
CoderMChu, line 434: SnomedComponent<?> should be ReferenceSetMember type

CoderMChu (commented):
@dmcgihtsdo It looks ok apart from minor changes. Thanks.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
dmcgihtsdo: Sorry, I never noticed you were talking about the `joinAxiomReferenceSetMembers` method @CoderMChu. Good catch. I'll update this method.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
dmcgihtsdo: @CoderMChu, I've had to use `?` because I am re-using the `mapByIdentifier` method for all components. This method stores a list of components in a map where the key is the component's identifier. 

Shall I create several non-generic `mapByIdentifier` methods or keep as is?

dmcgihtsdo: I have updated the methods to make better use of generics to prevent casting.

src/main/java/org/snomed/snowstorm/core/data/services/AutoMerger.java
dmcgihtsdo: I have made this change.

CoderMChu (approved):
It looks good now. Thanks.

PR #481, opened 2023-02-08
FRI-600 Indicate whether RHS of conflict window is version behind

dmcgihtsdo:
FRI-600 is concerned with indicating whether the RHS of the conflict window is a version behind, i.e. the LHS is February 2023 but the RHS is January 2023.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/domain/review/MergeReviewConceptVersions.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/domain/review/MergeReviewConceptVersions.java
@@ -8,6 +8,7 @@ public class MergeReviewConceptVersions {
 	private Concept targetConcept;
 	private Concept autoMergedConcept;
 	private Concept manuallyMergedConcept;
+	private boolean targetConceptVersionBehind;
 
 	public MergeReviewConceptVersions(Concept sourceConcept, Concept targetConcept) {
 		this.sourceConcept = sourceConcept;
@@ -46,6 +47,14 @@ public class MergeReviewConceptVersions {
 		this.manuallyMergedConcept = manuallyMergedConcept;
 	}
 
+	public boolean isTargetConceptVersionBehind() {
+		return targetConceptVersionBehind;
+	}
+
+	public void setTargetConceptVersionBehind(boolean targetConceptVersionBehind) {
+		this.targetConceptVersionBehind = targetConceptVersionBehind;
+	}
+
 	@Override
 	public String toString() {
 		return "MergeReviewConceptVersions{" +
@@ -53,6 +62,7 @@ public class MergeReviewConceptVersions {
 				", targetConcept=" + targetConcept +
 				", autoMergedConcept=" + autoMergedConcept +
 				", manuallyMergedConcept=" + manuallyMergedConcept +
+				", targetConceptDiverged=" + targetConceptVersionBehind +
 				'}';
 	}
 }
--- a/src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import io.kaicode.elasticvc.api.BranchService;
+import io.kaicode.elasticvc.api.PathUtil;
 import io.kaicode.elasticvc.api.VersionControlHelper;
 import io.kaicode.elasticvc.domain.Branch;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
@@ -52,6 +53,9 @@ public class BranchReviewService {
 	private BranchService branchService;
 
 	@Autowired
+	private SBranchService sBranchService;
+
+	@Autowired
 	private VersionControlHelper versionControlHelper;
 
 	@Autowired
@@ -80,6 +84,9 @@ public class BranchReviewService {
 
 	@Autowired
 	private AutoMerger autoMerger;
+
+	@Autowired
+	private CodeSystemService codeSystemService;
 
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
@@ -147,7 +154,7 @@ public class BranchReviewService {
 		assertMergeReviewCurrent(mergeReview);
 
 		final Set<Long> conceptsChangedInBoth = getConflictingConceptIds(mergeReview);
-
+		boolean targetBranchVersionBehind = isTargetBranchVersionBehind(mergeReview);
 		final Map<Long, MergeReviewConceptVersions> conflicts = new HashMap<>();
 		if (!conceptsChangedInBoth.isEmpty()) {
 
@@ -173,12 +180,75 @@ public class BranchReviewService {
 					if (sourceVersion != null && targetVersion != null) {
 						// Neither deleted, auto-merge.
 						mergeVersion.setAutoMergedConcept(autoMerger.autoMerge(sourceVersion, targetVersion, mergeReview.getTargetPath()));
+						if (sourceVersion.isReleased()) {
+							mergeVersion.setTargetConceptVersionBehind(targetBranchVersionBehind);
+						}
 					}
 					conflicts.put(conceptId, mergeVersion);
 				}
 			});
 		}
 		return conflicts.values();
+	}
+
+	private boolean isTargetBranchVersionBehind(MergeReview mergeReview) {
+		String sourcePath = mergeReview.getSourcePath();
+		String targetPath = mergeReview.getTargetPath();
+		Branch originalSourceBranch = getOriginatingTopLevelBranch(sourcePath);
+		Branch originalTargetBranch = getOriginatingTopLevelBranch(targetPath);
+
+		long leftTopLevelBaseTimestamp = originalSourceBranch.getBaseTimestamp();
+		long rightTopLevelBaseTimestamp = originalTargetBranch.getBaseTimestamp();
+		if (rightTopLevelBaseTimestamp < leftTopLevelBaseTimestamp) {
+			return true;
+		}
+
+		long leftTopLevelHeadTimestamp = originalSourceBranch.getHeadTimestamp();
+		long rightTopLevelHeadTimestamp = originalTargetBranch.getHeadTimestamp();
+		if (leftTopLevelHeadTimestamp == rightTopLevelHeadTimestamp) {
+			return false;
+		}
+
+		CodeSystem codeSystem = codeSystemService.findClosestCodeSystemUsingAnyBranch(sourcePath, false);
+		if (codeSystem == null) {
+			return false;
+		}
+
+		Branch lhsReleaseBranch = codeSystemService.findVersionBranchByCodeSystemAndBaseTimestamp(codeSystem, leftTopLevelHeadTimestamp);
+		if (lhsReleaseBranch == null) {
+			return false;
+		}
+
+		return lhsReleaseBranch.getHeadTimestamp() > rightTopLevelHeadTimestamp;
+	}
+
+	private Branch getOriginatingTopLevelBranch(String path) {
+		if (isCodeSystemBranch(path)) {
+			return branchService.findBranchOrThrow(path);
+		}
+
+		Branch originatingParentBranch = null;
+		String tmpPath = path;
+		Branch tmpBranch = branchService.findLatest(tmpPath);
+		boolean findingTopLevelBaseTimestamp = true;
+		while (findingTopLevelBaseTimestamp) {
+			String parentPath = PathUtil.getParentPath(tmpPath);
+			Branch parentBranch = sBranchService.findByPathAndHeadTimepoint(parentPath, tmpBranch.getBase().getTime());
+
+			if (isCodeSystemBranch(parentPath)) {
+				findingTopLevelBaseTimestamp = false;
+				originatingParentBranch = parentBranch;
+			}
+
+			tmpPath = parentPath;
+			tmpBranch = parentBranch;
+		}
+
+		return originatingParentBranch;
+	}
+
+	private boolean isCodeSystemBranch(String path) {
+		return path.equals("MAIN") || path.startsWith("SNOMEDCT-", path.lastIndexOf("/") + 1);
 	}
 
 	public void applyMergeReview(MergeReview mergeReview) throws ServiceException {
--- a/src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
@@ -40,6 +40,7 @@ import org.springframework.data.util.Pair;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.stereotype.Service;
 
+import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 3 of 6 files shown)


src/main/java/org/snomed/snowstorm/core/data/domain/review/MergeReviewConceptVersions.java
```diff
@@ -8,6 +8,7 @@ public class MergeReviewConceptVersions {
 	private Concept targetConcept;
 	private Concept autoMergedConcept;
 	private Concept manuallyMergedConcept;
+	private boolean targetConceptVersionBehind;
```
dmcgihtsdo, line 11: If the RHS is behind a version, then this boolean will be set to true.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -178,6 +179,29 @@ public Collection<MergeReviewConceptVersions> getMergeReviewConflictingConcepts(
 		return conflicts.values();
 	}
 
+	private boolean isTargetConceptVersionBehind(Concept sourceVersion, Concept targetVersion) {
+		Set<Integer> sourceEffectiveTimes = new HashSet<>();
+		sourceEffectiveTimes.add(sourceVersion.getReleasedEffectiveTime());
+		sourceEffectiveTimes.addAll(sourceVersion.getRelationships().stream().map(Relationship::getReleasedEffectiveTime).collect(Collectors.toList()));
+		sourceEffectiveTimes.addAll(sourceVersion.getDescriptions().stream().map(Description::getReleasedEffectiveTime).collect(Collectors.toList()));
```
dmcgihtsdo, line 186: Collect all release versions on the LHS and RHS. 

If the RHS does not contain all of the LHS, then it has fallen behind.

dmcgihtsdo:
This ticket is to update the API. The UI will pick up the new property in FRI-602.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
CoderMChu: @dmcgihtsdo You have left out language refset components.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -178,6 +179,29 @@ public Collection<MergeReviewConceptVersions> getMergeReviewConflictingConcepts(
...
+		sourceEffectiveTimes.addAll(sourceVersion.getDescriptions().stream().map(Description::getAssociationTargetMembers).flatMap(List::stream).map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+		sourceEffectiveTimes.addAll(sourceVersion.getAllOwlAxiomMembers().stream().map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+		sourceEffectiveTimes.addAll(sourceVersion.getInactivationIndicatorMembers().stream().map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+		sourceEffectiveTimes.addAll(sourceVersion.getAssociationTargetMembers().stream().map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+		while (sourceEffectiveTimes.remove(null)) ;
```
CoderMChu, line 191: Not sure why do we need while here.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -178,6 +179,29 @@ public Collection<MergeReviewConceptVersions> getMergeReviewConflictingConcepts(
...
+
+		Set<Integer> targetEffectiveTimes = new HashSet<>();
+		targetEffectiveTimes.add(targetVersion.getReleasedEffectiveTime());
+		targetEffectiveTimes.addAll(targetVersion.getRelationships().stream().map(Relationship::getReleasedEffectiveTime).collect(Collectors.toList()));
+		targetEffectiveTimes.addAll(targetVersion.getDescriptions().stream().map(Description::getReleasedEffectiveTime).collect(Collectors.toList()));
```
CoderMChu, line 196: The targetEffectiveTimes is a set already so we could collect them as set instead of list. Also we can add isReleased() filter to avoid null values.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -178,6 +179,29 @@ public Collection<MergeReviewConceptVersions> getMergeReviewConflictingConcepts(
...
+		targetEffectiveTimes.addAll(targetVersion.getDescriptions().stream().map(Description::getAssociationTargetMembers).flatMap(List::stream).map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+		targetEffectiveTimes.addAll(targetVersion.getAllOwlAxiomMembers().stream().map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+		targetEffectiveTimes.addAll(targetVersion.getInactivationIndicatorMembers().stream().map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+		targetEffectiveTimes.addAll(targetVersion.getAssociationTargetMembers().stream().map(ReferenceSetMember::getReleasedEffectiveTime).collect(Collectors.toList()));
+
+		return !targetEffectiveTimes.containsAll(sourceEffectiveTimes);
```
CoderMChu, line 202: Comparing released effective times will work for most of time but in certain cases this logic will not work. Especially  when international releases start releasing on the first day of every month. For example US edition 20230301 is dependant on international 20230131 release.  We start releasing on the first day of each month for international. After upgrading US extension to international 20230701, during rebase source and target might have same set of effective times however there might be changes made for international components in 20230301. This method will return false but in fact the target is a version behind.

CoderMChu (commented):
@dmcgihtsdo I have added a couple of comments.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
dmcgihtsdo: Wow, good edge case @CoderMChu! I'll work on improving this logic.

dmcgihtsdo: I have updated the logic @CoderMChu. Now, only components in the International module will have their released effective time read. If the US only has 2022-01-31 International releasedEffectiveTimes, then they have diverged.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
dmcgihtsdo: Thanks for the suggestion @CoderMChu. I have updated the method and done this.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
dmcgihtsdo: This was to remove null values. I have tweaked the code to only add versions if the component is released.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
CoderMChu: @dmcgihtsdo Does this new API need to check extension content is a version behind? If it does then above fix won't work.

dmcgihtsdo: It does need to check Extensions @CoderMChu. I'm not sure why you think it won't work; let's have a call?

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
```diff
@@ -522,12 +522,18 @@ public CodeSystemVersion findLatestVisibleVersion(String shortName) {
...
-						.withQuery(boolQuery().should(rangeQuery(CodeSystemVersion.Fields.IMPORT_DATE)
-								.lte(epochToImportDate(upperBoundEpoch))
-								.gte(epochToImportDate(lowerBoundEpoch))))
+						.withQuery(
+								boolQuery()
+										.must(
+												rangeQuery(CodeSystemVersion.Fields.IMPORT_DATE).lte(epochToImportDate(upperBoundEpoch)).gte(epochToImportDate(lowerBoundEpoch))
+										)
+										.must(
+												matchQuery(CodeSystemVersion.Fields.SHORT_NAME, shortName)
```
CoderMChu, line 534: Here is better to use termQuery as shot name field is mapped as keyword type.

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -3630,6 +3644,222 @@ void testTogglingInactivatingDescription() throws ServiceException {
...
+		concept = conceptService.create(
+				new Concept()
+						.addDescription(new Description("Extension maintained module").setTypeId(FSN).setCaseSignificance(ci).setAcceptabilityMap(intPreferred))
+						.addDescription(new Description("Extension maintained").setTypeId(SYNONYM).setCaseSignificance(ci).setAcceptabilityMap(intPreferred))
+						.addDescription(new Description("Extension nrc").setTypeId(SYNONYM).setCaseSignificance(ci).setAcceptabilityMap(intAcceptable))
+						.addAxiom(new Relationship(ISA, MODULE)),
+				extMain
+		);
+		String extModule = concept.getConceptId();
+		branchService.updateMetadata(extMain, Map.of(Config.DEFAULT_MODULE_ID_KEY, extModule));
```
CoderMChu, line 3784: Good test coverage.

CoderMChu (commented):
@dmcgihtsdo It looks better now. Thanks.

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
dmcgihtsdo: I have made this change.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -178,6 +188,58 @@ public Collection<MergeReviewConceptVersions> getMergeReviewConflictingConcepts(
...
+
+		long timestamp = 0;
+		String tmpPath = path;
+		Branch tmpBranch = branchService.findLatest(tmpPath);
+		boolean findingTopLevelBaseTimestamp = true;
+		while (findingTopLevelBaseTimestamp) {
+			String parentPath = PathUtil.getParentPath(tmpPath);
+			Branch parentBranch = sBranchService.findByPathAndHeadTimepoint(parentPath, tmpBranch.getBase().getTime());
+
+			isCodeSystemBranch = parentPath.equals("MAIN") || parentPath.startsWith("SNOMEDCT-", parentPath.lastIndexOf("/") + 1);
```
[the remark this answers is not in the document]
dmcgihtsdo: I have made this change.

dmcgihtsdo:
@CoderMChu, I have implemented your suggestions. I have noticed though that there are two flaky tests: testAutoMergeWhenDescriptionDeletedSomewhereAndVersionedElsewhere and testAutoMergeWhenAxiomDeletedSomewhereAndVersionedElsewhere. 

These tests will sometimes pass or sometimes fail. Can you double check my logic for `isTargetBranchVersionBehind` and `findByImportDateBetweenRange`? From what I can see, these two tests will fail because of a 1 second gap between MAIN's head and the time the version was created.

src/main/java/org/snomed/snowstorm/core/data/services/CodeSystemService.java
```diff
@@ -622,4 +642,16 @@ public void updateCodeSystemBranchMetadata(CodeSystem codeSystem) {
...
+		Date date = new Date(epoch);
+		date.setTime(date.getTime() - 5000);
+		return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss").format(date);
+	}
+
+	private String epochToImportDatePlusTime(long epoch, long plusTime) {
+		Date date = new Date(epoch);
+		date.setTime(date.getTime() + 5000);
+		return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss").format(date);
+	}
```
dmcgihtsdo, line 656: @CoderMChu, I'm not 100% on this inflation (+/- 5 seconds) logic going to develop. Without it, a couple of tests will occasionally fail however. 

I'm not sure whether the failing tests themselves are at fault or they've helped expose a slight gap in our logic (there are a couple seconds overlay in versioning vs the heads of branches).

dmcgihtsdo: Scenario A
- version has an import time of 09:00:03
- main has a head time of 09:00:03
- task has a head time of 09:00:01

In this scenario, without the inflation, the tests pass.

Scenario B
- version has an import time of 09:00:05
- main has a head time of 09:00:03
- task has a head time of 09:00:01

In this scenario, without the inflation, the test fails. That is why I have added a +/- 5 seconds.

dmcgihtsdo: In theory, there shouldn't be a CodeSystem with multiple versions within this inflation... it just seems messy.

src/main/java/org/snomed/snowstorm/core/data/services/BranchReviewService.java
```diff
@@ -203,14 +203,18 @@ private boolean isTargetBranchVersionBehind(MergeReview mergeReview) {
...
+		}
+
+		Branch releaseBranch = sBranchService.findLatestReleaseBranchByCodeSystem(codeSystem);
+		if (releaseBranch == null) {
+			return false;
 		}
 
-		return !codeSystemService.findByImportDateBetweenRange(codeSystemShortName, leftTopLevelHeadTimestamp, rightTopLevelHeadTimestamp).isEmpty();
+		// If no versions exist between the date range, then the target Branch is not a version behind.
+		return !codeSystemService.findByImportDateBetweenRange(codeSystem.getShortName(), releaseBranch.getHeadTimestamp(), rightTopLevelHeadTimestamp).isEmpty();
```
CoderMChu, line 217: @dmcgihtsdo After use MAN's head time to look up the release branch (base's time), we don't need to check code system version. We can just compare these two dates.

CoderMChu (commented):
@dmcgihtsdo As discussed that using the latest release branch is not quite right. Your alternative solution should work. Thanks.

CoderMChu (commented):
Yes it looks good apart from minor change.

src/main/java/org/snomed/snowstorm/core/data/services/SBranchService.java
```diff
@@ -119,6 +119,25 @@ public List<Branch> findByPathAndBaseTimepoint(Set<String> path, Date baseTimest
 				.stream().map(SearchHit::getContent).collect(Collectors.toList());
 	}
 
+	public Branch findByPathAndHeadTimepoint(String path, long head) {
+		SearchHits<Branch> query = elasticsearchTemplate.search(
+				new NativeSearchQueryBuilder()
+						.withQuery(
+								boolQuery()
+										.must(matchQuery("path", path))
```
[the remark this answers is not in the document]
dmcgihtsdo: Thanks @CoderMChu, I have made this change.

dmcgihtsdo:
@CoderMChu has given me the thumbs up privately.

PR #484, opened 2023-02-17
RT2-1567 - refset multibranch searching

RichardAWood:
Switching multibranch description search to search REFSET branches also, instead of just published branches.  Also switching from passing in ECL to specifying "RefsetsOnly", and hard-coding "<446609009" as the ECL.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
@@ -31,6 +31,8 @@ import org.springframework.util.CollectionUtils;
 
 import java.time.LocalDate;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static io.kaicode.elasticvc.api.ComponentService.LARGE_PAGE;
@@ -53,6 +55,9 @@ public class MultiSearchService implements CommitListener {
 	private ECLQueryService eclQueryService;
 
 	@Autowired
+	private BranchService branchService;
+
+	@Autowired
 	private CodeSystemService codeSystemService;
 
 	@Autowired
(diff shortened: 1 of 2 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
```diff
@@ -68,26 +73,37 @@ public class MultiSearchService implements CommitListener {
...
 			// ECL -> conceptIds
-			BranchCriteria branchCriteria = versionControlHelper.getBranchCriteria("MAIN");
-			//MultiBranchCriteria branchesQuery = getBranchesQuery();
-			Page<Long> page = eclQueryService.selectConceptIds(ecl, branchCriteria, true, null, null);
-			criteria.conceptIds(page.getContent());
+			MultiBranchCriteria branchesQuery = getBranchesQueryWithRefsets();
+			Set<Long> conceptIds = new HashSet<>();
+			for(BranchCriteria branchCriteria : branchesQuery.getBranchCriteria())
+			{
+				conceptIds.addAll(eclQueryService.selectConceptIds("<446609009", branchCriteria, true, null, null).getContent());
```
pgwilliams, line 91: Please add this SCTID to org.snomed.snowstorm.core.data.domain.Concepts and reference as a String constant.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
```diff
@@ -229,6 +288,42 @@ public Set<String> getAllPublishedVersionBranchPaths() {
 		return publishedVersionBranchPaths;
 	}
 	
+	public Set<String> getCurrentRefsetBranchPaths() {
+		List<Branch> allBranches = branchService.findAll();
+		Set<String> refsetBranchPaths = new HashSet<>();
+		Map<String,String> refsetToCurrentBranchPathMap = new HashMap<>();
+		//Identify the latest Refset edit branch for each refset
+		//For example:
+		//MAIN/SNOMEDCT/REFSETS/REFSET-48353439008-1676579557135/EDIT-1676579564596
```
pgwilliams, line 297: Does this work by each commit being made on a uniquely named sub-branch?   This is sort of how we work for authoring, although we keep the same path name and just create a new object.   So the previous one gets given an 'end' timestamp which is equal to the time stamp of the next object.    Any concerns here @kaicode ?

So can you have multiple edits that are all "live" at the same time?   What happens to the 'end' parameter on the branch objects during successive edits?

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
```diff
@@ -229,6 +288,42 @@ public Set<String> getAllPublishedVersionBranchPaths() {
...
+		//Is for refset 4853439008, timestamp 1676579564596
+		final Pattern p = Pattern.compile("^(.*)(REFSET-)([0-9]+)(-[0-9]+\\/EDIT-)([0-9]+)$");
```
pgwilliams, line 299: This is extremely client-application specific to be putting into a terminology server.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
```diff
@@ -68,26 +73,37 @@ public class MultiSearchService implements CommitListener {
 	private MultiBranchCriteria cachedBranchCriteria = null;
 	private LocalDate cacheDate = null;
 
-	public Page<Description> findDescriptions(DescriptionCriteria criteria, String ecl, PageRequest pageRequest) {
-
-		if (ecl != null) {
+	public Page<Description> findDescriptions(DescriptionCriteria criteria, Boolean refsetsOnly, PageRequest pageRequest) {
```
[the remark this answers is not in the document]
RichardAWood: Yep, that's correct.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
RichardAWood: Ah, I hadn't discovered the Concepts section yet. And actually, it looks like it's already in there:

`public static final String REFSET_SIMPLE = "446609009";`

Will update this section of code to reference it.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
RichardAWood: You're not wrong.  But since this functionality is only accessible via the "refsetsOnly" parameter, having them reference the Refset tool specific branch patterns hopefully isn't too horrible.

I'm trying to think of other alternatives.  The problem we're trying to solve is that the existing MultiSearchService.findDescriptions functionality only searches published branches, and one of the Refset Tool requirements is to be able to search for refsets by description across all branches, including unpublished ones.  An alternative to hard coding the regex could be to expose something like "include branches" parameter in the api, and the refset tool could grab all branches, parse them, and pass the ones it wants in. But each branch is a long string and we would need to pass in _every_ in-progress refset branch, which would make a monster call...

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
```diff
@@ -229,6 +288,42 @@ public Set<String> getAllPublishedVersionBranchPaths() {
...
+		//Identify the latest Refset edit branch for each refset
+		//For example:
+		//MAIN/SNOMEDCT/REFSETS/REFSET-48353439008-1676579557135/EDIT-1676579564596
+		//Is for refset 4853439008, timestamp 1676579564596
+		final Pattern p = Pattern.compile("^(.*)(REFSET-)([0-9]+)(-[0-9]+\\/EDIT-)([0-9]+)$");
+		synchronized(this) {
+			for (Branch b : allBranches) {
+				Matcher m = p.matcher(b.getPath());
+				if (m.find()) {
+					final String refset = m.group(3);
```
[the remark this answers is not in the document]
RichardAWood: Your suggestion is a good one.  Creating 'final int's  in the "getCurrentRefsetBranchPaths" method rather than 'final static int's at the top, since they're only used in this one place.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
RichardAWood: Updates made and changes pushed.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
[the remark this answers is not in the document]
RichardAWood: Updates made and changes pushed.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
twkatadin: Each refset that is worked on gets it's own refset specifc branch with a timestamp when the edit cycle is begun. Under that branch there is a timestamped edit branch created for every time the refset is placed into "In Edit" mode. This can occur multiple times during an edit cycle. We do this since we can't delete branches but still need a way to undo changes while retaining the ability to recover undone edits in the rare case a massive change list was mistakenly cancelled. Each refset specific branch will only ever have one active edit branch, which is the one with the latest timestamp. This process was reviewed by Rory.

src/main/java/org/snomed/snowstorm/core/data/services/MultiSearchService.java
pgwilliams: Sounds like you've some options there.  Like, I liked the ecl parameter because I said "oh that's something that people might be generally useful", but this code is so specific to the internal workings of your application that it can't be used anywhere else.    Have you time for a call in the next couple of hours @RichardAWood ?

RichardAWood: Hi Peter,
I'm online now, and have a call in 20 minutes (8:00-8:30 am PST) - want to try to sneak in a talk before or after that?

PR #485, opened 2023-02-17
Use elasticksearch:7.17.9 in docker-compose.yml

adamkover-felleskatalogen:
snowstorm:latest is now version 8, no longer compatible with elasticsearch:7.7.0

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '2.1'
 services:
   elasticsearch:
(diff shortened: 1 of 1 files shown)


CoderMChu:
@adamkover-felleskatalogen Yes that's correct. The Elasticsearch version need to be between 7.10.0 and 7.17.9. I forgot to update the docker file in version 8 pre-release. We are using 7.10.2 in production and will update this in the coming release. Thanks.

PR #489, opened 2023-02-28
RT2-1567: Creating POST version of multisearch findDescriptions 

RichardAWood:
The intention is to allow users to specify additional, non-published branches to include in the search.  This will be used by RT2 to search refset names across all branches, and can potentially be useful to other others as well.


pgwilliams (approved):
Yeah, looks really good.   Nice and generic, with the potentially to be more widely useful.
Thanks for reworking that.

PR #519, opened 2023-06-05
ISTO-94 Snowstorm FHIR Inctive Flag

kaicode:
This PR contains various fixes for the Snowstorm FHIR interface relating to the inactive flag.
- Fixed ValueSet $expand
  - inactive flag was inverted)
- Fixed CodeSystem $lookup
  - Added "inactive" property to conform with standard
  - Removed "active" property because redundant
- Added "inactive" property to CodeSystem $validate-code
- Added "inactive" property to ValueSet $validate-code


pgwilliams (approved):
Yeah really nice code and great spotting on the fact that our logic was reversed there!   Good to get that fixed ASAP.   Also appreciate the unit tests.

PR #520, opened 2023-06-05
ISTO-95 Block FULL import unless on empty MAIN.

kaicode:
This PR stops anyone using the FULL import on branches other than MAIN. The branch also needs to be empty. This block should have been there from the start because all content is imported against the current parent code system version, this is incorrect behaviour.
A complete child Edition/Extension import feature could be implemented in the future.


CoderMChu (approved):
@kaicode It looks good. Thanks

PR #525, opened 2023-06-23
ISTO-97 defaultModule for desc inactivation indicator

kaicode:
This is a bug fix to set the module for description inactivation indicators on new descriptions on already inactive concepts. Before this fix the module was blank.


pgwilliams (approved):
Lovely unit test.   Will merge now.

PR #529, opened 2023-07-07
RT2-1567: Creating POST version of multisearch findDescriptions to allow users to specify additional, non-published branches to include in the search

njmarques:


njmarques:
Replaced with PR 530.

PR #545, opened 2023-08-28
PIP-333 Add ability to locally persist export

dmcgihtsdo:
PIP-333 is concerned with adding the ability for the UI to download an RF2 export without timing out. To do this, I have added two things:

- Status to ExportConfiguration for polling
- Option to export archive locally

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
@@ -3,6 +3,8 @@ package org.snomed.snowstorm.core.data.domain.jobs;
 import io.swagger.v3.oas.annotations.media.Schema;
 import org.snomed.snowstorm.core.rf2.RF2Type;
 import org.springframework.data.elasticsearch.annotations.Document;
+import org.springframework.data.elasticsearch.annotations.Field;
+import org.springframework.data.elasticsearch.annotations.FieldType;
 
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
@@ -47,6 +49,16 @@ public class ExportConfiguration {
 
 	@Schema(description = "If refsetIds are included, this indicates that the export will be a refset-only export.")
 	private Set<String> refsetIds;
+
+	@Schema(defaultValue = "PENDING")
+	private ExportStatus status;
+
+	@Schema
+	@Field(type = FieldType.Keyword)
+	private String exportFilePath;
+
+	@Schema(defaultValue = "false")
+	private boolean startExport;
 
 	public ExportConfiguration() {
 	}
@@ -150,5 +162,29 @@ public class ExportConfiguration {
 
 	public void setRefsetIds(Set<String> refsetIds) {
 		this.refsetIds = refsetIds;
+	}
+
+	public ExportStatus getStatus() {
+		return status;
+	}
+
+	public void setStatus(ExportStatus status) {
+		this.status = status;
+	}
+
+	public String getExportFilePath() {
+		return exportFilePath;
+	}
+
+	public void setExportFilePath(String exportFilePath) {
+		this.exportFilePath = exportFilePath;
+	}
+
+	public boolean isStartExport() {
+		return startExport;
+	}
+
+	public void setStartExport(boolean startExport) {
+		this.startExport = startExport;
 	}
 }
--- a/src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportStatus.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportStatus.java
@@ -1,0 +1,9 @@
+package org.snomed.snowstorm.core.data.domain.jobs;
+
+public enum ExportStatus {
+    PENDING,
+    RUNNING,
+    COMPLETED,
+    DOWNLOADED,
+    FAILED
+}
--- a/src/main/java/org/snomed/snowstorm/core/rf2/export/ExportService.java
+++ b/src/main/java/org/snomed/snowstorm/core/rf2/export/ExportService.java
@@ -16,6 +16,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.snomed.snowstorm.core.data.domain.*;
 import org.snomed.snowstorm.core.data.domain.jobs.ExportConfiguration;
+import org.snomed.snowstorm.core.data.domain.jobs.ExportStatus;
 import org.snomed.snowstorm.core.data.repositories.ExportConfigurationRepository;
 import org.snomed.snowstorm.core.data.services.BranchMetadataHelper;
 import org.snomed.snowstorm.core.data.services.BranchMetadataKeys;
@@ -38,6 +39,7 @@ import org.springframework.util.CollectionUtils;
 import java.io.*;
 import java.nio.file.Files;
 import java.util.*;
+import java.util.concurrent.ExecutorService;
 import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
@@ -73,6 +75,9 @@ public class ExportService {
 	@Autowired
 	private CodeSystemService codeSystemService;
 
+	@Autowired
+	private ExecutorService executorService;
+
 	private final Set<String> refsetTypesRequiredForClassification = Sets.newHashSet(Concepts.REFSET_MRCM_ATTRIBUTE_DOMAIN, Concepts.OWL_EXPRESSION_TYPE_REFERENCE_SET);
 
 	private final Logger logger = LoggerFactory.getLogger(getClass());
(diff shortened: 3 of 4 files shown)


src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
```diff
@@ -48,6 +48,12 @@ public class ExportConfiguration {
 	@Schema(description = "If refsetIds are included, this indicates that the export will be a refset-only export.")
 	private Set<String> refsetIds;
 
+	@Schema(defaultValue = "PENDING")
+	private String status;
```
dmcgihtsdo, line 52: Whenever an export job is created, it will have the "PENDING" status.

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
```diff
@@ -48,6 +48,12 @@ public class ExportConfiguration {
...
+
+	@Schema(defaultValue = "false")
+	private boolean persist;
+
```
dmcgihtsdo, line 56: This flag controls whether the export should be written to a local "exports" folder.

src/main/java/org/snomed/snowstorm/core/rf2/export/ExportService.java
```diff
@@ -118,15 +133,30 @@ public void exportRF2Archive(ExportConfiguration exportConfiguration, OutputStre
...
+			exportConfiguration.setStatus(ExportStatus.FAILED);
+			exportConfigurationRepository.save(exportConfiguration);
 			throw new ExportException("Failed to copy RF2 data into output stream.", e);
 		} finally {
 			exportFile.delete();
 			logger.info("Deleted " + exportConfiguration.getId() + " export file " + exportFile);
 		}
 	}
 
+	public void copyRF2Archive(ExportConfiguration exportConfiguration, OutputStream outputStream) throws ExportException {
```
dmcgihtsdo, line 149: This method copies the local archive into the request's output stream.

src/main/java/org/snomed/snowstorm/rest/ExportController.java
```diff
@@ -38,12 +39,19 @@ public class ExportController {
...
+
 	@Operation(summary = "Create an export job.",
 			description = "Create a job to export an RF2 archive. " +
 					"The 'location' response header contain the URL, including the identifier, of the new resource.")
 	@PostMapping
-	public ResponseEntity<Void> createExportJob(@Valid @RequestBody ExportRequestView exportConfiguration) {
-		String id = exportService.createJob(exportConfiguration);
+	public ResponseEntity<Void> createExportJob(@Valid @RequestBody ExportRequestView exportRequestView) throws InterruptedException {
+		String id = exportService.createJob(exportRequestView);
+		if (exportRequestView.isPersist()) {
```
dmcgihtsdo, line 51: If `persist` is set to true, the export will automatically start. By default it is false so there should be no regressions.

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
CoderMChu: @dmcgihtsdo We can use ExportStatus as the data type.

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
```diff
@@ -48,6 +48,15 @@ public class ExportConfiguration {
 	@Schema(description = "If refsetIds are included, this indicates that the export will be a refset-only export.")
 	private Set<String> refsetIds;
 
+	@Schema(defaultValue = "PENDING")
+	private String status;
+
+	@Schema
+	private String exportPath;
```
CoderMChu, line 55: @dmcgihtsdo I assume this is the exported file path. Can you rename it to exportFilepath please? Also we need consider that these files will be deleted during system restart. If they haven't been downloaded already then we need ask users to start a new export.

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
```diff
@@ -48,6 +48,15 @@ public class ExportConfiguration {
...
+
+	@Schema(defaultValue = "false")
+	private boolean startExport;
```
CoderMChu, line 58: @dmcgihtsdo As discussed before we shouldn't really need this workaround if we have made existing apis backward compatible. If this is a must then please make this transient as we don't want this to be persisted.

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
CoderMChu: Also can you add @Field(type = FieldType.Keyword) for this please? This field is only useful if we store exported file in S3.

src/main/java/org/snomed/snowstorm/rest/ExportController.java
```diff
@@ -38,12 +45,19 @@ public class ExportController {
...
+    private ExecutorService executorService;
+
 	@Operation(summary = "Create an export job.",
 			description = "Create a job to export an RF2 archive. " +
 					"The 'location' response header contain the URL, including the identifier, of the new resource.")
 	@PostMapping
-	public ResponseEntity<Void> createExportJob(@Valid @RequestBody ExportRequestView exportConfiguration) {
-		String id = exportService.createJob(exportConfiguration);
+	public ResponseEntity<Void> createExportJob(@Valid @RequestBody ExportRequestView exportRequestView) {
+		String id = exportService.createJob(exportRequestView);
```
CoderMChu, line 56: @dmcgihtsdo It is a good idea to use ExecutorService. Can you move these logic into the ExportService please?

src/main/java/org/snomed/snowstorm/rest/ExportController.java
```diff
@@ -59,10 +73,33 @@ public ExportConfiguration getExportJob(@PathVariable String exportId) {
...
+			if (Objects.equals(ExportStatus.DISCARDED, exportConfiguration.getStatus())) {
+				response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+				response.getWriter().write("Archive has already been downloaded; cannot re-download");
+				response.getWriter().flush();
+				return;
+			}
+
+			// Initial check before polling
+			if (Objects.equals(ExportStatus.PENDING, exportConfiguration.getStatus()) || Objects.equals(ExportStatus.IN_PROGRESS, exportConfiguration.getStatus())) {
+				exportConfiguration = pollExportStatus(exportConfiguration);
```
CoderMChu, line 91: @dmcgihtsdo If you insist to use the workaround (i.e with startExport flag) then we don't need to add backward compatibility here. You should respond with bad request when export is not completed.

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportStatus.java
```diff
@@ -0,0 +1,9 @@
+package org.snomed.snowstorm.core.data.domain.jobs;
+
+public enum ExportStatus {
+    PENDING,
+    IN_PROGRESS,
+    READY_FOR_DOWNLOAD,
```
CoderMChu, line 6: @dmcgihtsdo Can we have PENDING, RUNNING, COMPLETED,FAILED, DOWNLOADED please?

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportConfiguration.java
dmcgihtsdo: Thanks @CoderMChu, I have made those changes.

src/main/java/org/snomed/snowstorm/rest/ExportController.java
dmcgihtsdo: Thanks @CoderMChu, I have made this change. The method is now `exportRF2ArchiveAsync`.

src/main/java/org/snomed/snowstorm/rest/ExportController.java
dmcgihtsdo: Sorry but I'm not sure what you mean @CoderMChu. 

`exportRF2ArchiveAsync & copyRF2Archive` is backwards compatible, but for testing capacity, I've kept the use of `exportRF2Archive`.

src/main/java/org/snomed/snowstorm/core/data/domain/jobs/ExportStatus.java
dmcgihtsdo: Thanks @CoderMChu, I have made that change.

src/main/java/org/snomed/snowstorm/rest/ExportController.java
CoderMChu: @dmcgihtsdo We don't need to pollExportStatus() when you have used startExport flag for backward compatibility.  The correct API call should have checked the status before calling download api. (i.e The client should poll)

dmcgihtsdo: Thanks for the clarification @CoderMChu. I have made that change.

PR #553, opened 2023-09-13
ISTO-99 Snowstorm refset member API, add details to support SRS manifest generation

kaicode:
This PR updates the browser member API endpoint to add more details about each refset type. This is required to support generating SNOMED Release Service manifest xml files in the upcoming Simplex Tool.

I have also fixed the Snowstorm refset configuration for the ICD-10 and ICD-9-CM maps. The config now points to the generic Complex Map and Extended Map types rather than any specific refset. This is in line with the rest of the configuration and means that any other maps of these types will also pick up the correct configuration. This area of configuration is the source of the information that is now exposed in the refset API.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/ReferenceSetMemberService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/ReferenceSetMemberService.java
@@ -695,10 +695,15 @@ public class ReferenceSetMemberService extends ComponentService {
 				refsetToTypeMap.put(conceptId, conceptId);
 			} else {
 				String type = null;
-				for (Long ancestor : semanticIndexRefsetConcept.getAncestors()) {
-					if (refsetTypesFromConfig.contains(ancestor.toString())) {
-						// Refset ancestor matches one of the configured types. These are generally more specific than those in the hierarchy.
-						type = ancestor.toString();
+				if (refsetTypesFromConfig.contains(conceptId)) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/ReferenceSetMemberService.java
```diff
@@ -669,32 +669,53 @@ public PageWithBucketAggregations<ReferenceSetMember> findReferenceSetMembersWit
...
+				for (Long ancestor : semanticIndexRefsetConcept.getAncestors()) {
+					if (refsetTypesFromConfig.contains(ancestor.toString())) {
+						// Refset ancestor matches one of the configured types. These are generally more specific than those in the hierarchy.
+						type = ancestor.toString();
+					}
+				}
+				if (type == null) {
+					for (Long ancestor : semanticIndexRefsetConcept.getAncestors()) {
+						if (refsetTypesFromHierarchy.contains(ancestor)) {
+							// Refset ancestor matches on of the "type" refsets from the hierarchy
```
CoderMChu, line 712: @kaicode  We could add this as part of or condition in above for loop.

CoderMChu: @kaicode Please ignore my previous comment. I understand now that configured refset types are preferred. So it makes sense to process that first.

CoderMChu (approved):
@kaicode It looks good. Thanks for adding a nice and new feature.

PR #554, opened 2023-09-13
ISTO-86 Local Sequential SCTID Generator in Snowstorm

kaicode:
Snowstorm usually requires the Component Identifier Service to be deployed and configured as the source of truth for SNOMED SCTID Identifiers. This idea of that service was that many software components could generate ids for the same extension in a coordinated way without risk of id collision, by generating them in a central service.

An alternative approach is to always use Snowstorm to create new components and has it assign identifiers automatically based on the content already in Snowstorm. This PR enables that strategy and sets this internal sequential ID generator as the default for anyone using Snowstorm in a new setup. The SNOMED International Authoring Platform will continue to use the CIS service.


src/main/java/org/snomed/snowstorm/core/data/services/identifier/LocalSequentialIdentifierSource.java
```diff
@@ -0,0 +1,133 @@
...
+	 * Finds highest SCTID in the given namespace and partition across all branches and versions. Returns the sequence identifier.
+	 * @return Sequence from the highest existing SCTID.
+	 */
+	private int findHighestIdentifierSequence(int namespaceId, String partitionId) {
+
+		Class<? extends SnomedComponent<?>> componentClass = null;
+		String idField = null;
+
+		switch (partitionId) {
+			case "00":
```
CoderMChu, line 74: We can use Java 17 switch expression here now.

src/main/java/org/snomed/snowstorm/core/data/services/identifier/LocalSequentialIdentifierSource.java
```diff
@@ -0,0 +1,133 @@
...
+
+		SearchHits<? extends SnomedComponent<?>> searchHits = elasticsearchTemplate.search(new NativeSearchQueryBuilder()
+
+				// Regex query
+				.withQuery(regexpQuery(idField, regex))
+
+				// Numbers as strings require sorting by length and characters:
+				// Sort by string length first..
+				.withSort(new ScriptSortBuilder(new Script(String.format("doc['%s'].value.length()", idField)), ScriptSortBuilder.ScriptSortType.NUMBER).order(SortOrder.DESC))
+				// then sort by characters.
```
CoderMChu, line 120: Nice sorting strategy.

src/main/java/org/snomed/snowstorm/core/data/services/identifier/LocalSequentialIdentifierSource.java
```diff
@@ -86,7 +86,7 @@ private int findHighestIdentifierSequence(int namespaceId, String partitionId) {
 				componentClass = Relationship.class;
 				idField = Relationship.Fields.RELATIONSHIP_ID;
 				break;
-			case "06":
+			case "16":
```
CoderMChu, line 92: @kaicode Shouldn't it be "06" or "16" here? I can see POSTCOORDINATED_EXPRESSION_PARTITION_ID  is still set to "06".

src/main/java/org/snomed/snowstorm/core/data/services/identifier/LocalSequentialIdentifierSource.java
```diff
@@ -132,10 +135,17 @@ private int findHighestIdentifierSequence(int namespaceId, String partitionId) {
...
+				itemIdentifier = highestSctid.substring(0, highestSctid.length() - 3);
 			} else {
-				highestSequence = Integer.parseInt(highestSctid.substring(0, highestSctid.lastIndexOf(namespaceId + "")));
+				itemIdentifier = highestSctid.substring(0, highestSctid.lastIndexOf(namespaceId + ""));
+			}
+			if (itemIdentifier.length() <= 8) {
+				highestSequence = Integer.parseInt(itemIdentifier);
+			} else {
+				logger.error("{} found with bad SCTID '{}' on branch {}, this will prevent identifier generation.",
+						componentClass.getSimpleName(), highestSctid, component.getPath());
```
CoderMChu, line 148: It is a tricky scenario here. Assigning next sequence number to 100 if the SCTID found is invalid. It depends on the use case I think. Maybe snowstorm fails to start makes sense in some case because not everyone is checking the logs. Or maybe could we log error and switch to local random identifier when this happens?

CoderMChu (commented):
@kaicode I have added a couple of questions. Apart from that it looks good. Thanks.

src/main/java/org/snomed/snowstorm/core/data/services/identifier/LocalSequentialIdentifierSource.java
kaicode: I think this block was just to support unit testing where we had some badly made up data. I am comfortable not allowing ID generation to work correctly if there are bad SCTIDs in the data. 
The sequence number is being set of 100 for the International Edition to support unit testing when there is no data loaded, otherwise I think the generated ID is too short and doesn't match the spec. This would never happen because the International Edition would be loaded first and anyway it's maintained on the Managed Service using the CIS service.

kaicode:
@CoderMChu thanks for your review. I have fixed the switch statement and the POSTCOORDINATED_EXPRESSION_PARTITION_ID constant. 16 is the partition for expressions.

dmcgihtsdo:
Hi @kaicode, I was temporarily working on MAINT-2336 on develop which has ended up in this pr. Can you rebase/drop the commit from your branch please?

kaicode:
@dmcgihtsdo well spotted, I have now rebased to remove your commit.

src/main/java/org/snomed/snowstorm/core/data/services/identifier/LocalSequentialIdentifierSource.java
CoderMChu: As long as this is acceptable then logging error is fine by me.

CoderMChu (approved):

PR #559, opened 2023-10-02
Maint 2289 concept annotations

QuyenLy87:


src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.util.Objects;
+import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@JsonView(value = View.Component.class)
+public class Annotation extends ReferenceSetMember implements AnnotationView {
+
+	private static Pattern ANNOTATION_LANGUAGE_PATTERN = Pattern.compile("^@.*\"");
```
dmcgihtsdo, line 17: @QuyenLy87 this looks unused. What is it for? Can we remove it?

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,133 @@
...
+
+	private String annotationId;
+	private String annotationTypeId;
+
+	private ConceptMini annotationType;
+
+	private String annotationValue;
```
dmcgihtsdo, line 24: @QuyenLy87 can you remove the double spacings between the members?

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,133 @@
...
+	public String toString() {
+		return "Annotation{" +
+				"annotationId='" + annotationId + '\'' +
+				", annotationTypeId='" + annotationTypeId + '\'' +
+				", annotationType=" + annotationType +
+				", annotationValue='" + annotationValue + '\'' +
+				", annotationLanguage='" + annotationLanguage + '\'' +
+				", active=" + active +
+				'}';
+	}
```
dmcgihtsdo, line 132: @QuyenLy87 as well as the annotation specific properties, should we have the module, effective time etc in the toString method?

src/main/java/org/snomed/snowstorm/core/data/domain/Concept.java
```diff
@@ -240,6 +244,11 @@ public Concept addIdentifier(Identifier identifier) {
 		return this;
 	}
 
+	public Concept addAnnotation(Annotation annotation) {
+		annotations.add(annotation);
+		return this;
+	}
+
```
dmcgihtsdo, line 251: @QuyenLy87 this looks like it will throw a NPE depending on which constructor is used to instantiate the Concept.

QuyenLy87: Hi @dmcgihtsdo this is not the issue. The addAnnotation is similar to other methods (for example: addDescription, addRelationship, addAxiom...) When instantiating the concept, we always call the default constructor "public Concept()". So the property "annotations" will never be null.

QuyenLy87:
Hi @dmcgihtsdo I've made some changes regarding your feedback. Please review this PR again. 
Thanks

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -507,6 +507,28 @@ void testSaveConceptWithDescription() throws ServiceException {
 		assertEquals(1, description.getAcceptabilityMapFromLangRefsetMembers().size());
 	}
 
+	@Test
+	void testSaveConceptWithAnnotation() throws ServiceException {
+		final Concept concept = new Concept("50960005", 20020131, true, "900000000000207008", "900000000000074008");
+		concept.addDescription(new Description("84923010", 20020131, true, "900000000000207008", "50960005", "en", FSN,
```
dmcgihtsdo, line 513: @QuyenLy87 can you add a test for a rebasing/conflict scenario? There's quite a few in `BranchMergeServiceTest`. I wonder what will happen if we rebase a concept with an annotation against itself without an annotation. Depending on which column is selected, we shouldn't accidentally lose the annotation.

dmcgihtsdo:
@QuyenLy87 I created a Concept locally using this payload: 

`"annotations": [
      {
          "annotationId": "123",
          "annotationTypeId": "123",
          "annotationValue": "@en This is a new disease."
      }
  ],`

The response was: 

`"annotations": [
        {
            "deleted": false,
            "changed": false,
            "active": true,
            "released": false,
            "additionalFields": {},
            "annotationId": "123",
            "annotationTypeId": "123",
            "annotationValue": "@en This is a new disease."
        }
    ],`

`deleted`, `changed`, and `additionalFields` look like they shouldn't be in the response?

QuyenLy87:
Hi @dmcgihtsdo, I've added some test cases. Please have a look. Cheer

dmcgihtsdo:
Thanks for the rebase test @QuyenLy87. Can you add another one but this time with for a conflict scenario? Just to make sure we do/don't lose annotations as expected.

dmcgihtsdo:
@QuyenLy87, when trying to read a Concept with an annotation, it fell over with this stack:

`
java.lang.StringIndexOutOfBoundsException: begin 1, end -1, length 26
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4604)
	at java.base/java.lang.String.substring(String.java:2707)
	at org.snomed.snowstorm.core.data.domain.Annotation.fromRefsetMember(Annotation.java:88)
	at org.snomed.snowstorm.core.data.services.ConceptService.lambda$joinAnnotations$19(ConceptService.java:883)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
`

The annotation I created looks like this:

`
"annotations": [
      {
          "annotationId": "123",
          "annotationTypeId": "123",
          "annotationValue": "@en This is a new disease."
      }
  ],
`

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,137 @@
...
+import com.fasterxml.jackson.annotation.JsonView;
+import io.micrometer.core.instrument.util.StringUtils;
+import org.snomed.snowstorm.core.pojo.TermLangPojo;
+import org.snomed.snowstorm.rest.View;
+
+import java.util.Objects;
+import java.util.UUID;
+
+@JsonView(value = View.Component.class)
+@JsonIgnoreProperties({"memberId", "additionalFields", "effectiveTimeI", "deleted", "changed"})
```
dmcgihtsdo, line 14: Thanks for hiding these properties from the api response @QuyenLy87 however no other domain object seems to use the `JsonIgnoreProperties` annotation. Can you hide these properties without using the annotation so it is consistent across the domain?

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,137 @@
...
+public class Annotation extends ReferenceSetMember implements AnnotationView {
+
+	private String annotationId;
+	private String annotationTypeId;
+	private ConceptMini annotationType;
+	private String annotationValue;
+	private String annotationLanguage;
```
dmcgihtsdo, line 21: @QuyenLy87, going by the ticket, it looks like `annotationTarget` is missing. Do you know if this field is needed or is it an old requirement?

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,137 @@
...
+		setActive(fromMember.isActive());
+		setEffectiveTimeI(fromMember.getEffectiveTimeI());
+		setReferencedComponentId(fromMember.getReferencedComponentId());
+		setReleased(fromMember.isReleased());
+		setAnnotationTypeId(fromMember.getAdditionalField(AnnotationFields.ANNOTATION_TYPE_ID));
+
+		String annotationValue = fromMember.getAdditionalField(AnnotationFields.ANNOTATION_VALUE);
+		setAnnotationValue(annotationValue);
+		if (StringUtils.isNotEmpty(annotationValue) && annotationValue.startsWith("@")) {
+			setAnnotationLanguage(annotationValue.substring(1, annotationValue.indexOf("\"")));
```
dmcgihtsdo, line 88: @QuyenLy87, when creating a Concept with an annotation, the `annotationLanguage` property is not returned.

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
pgwilliams: We're discussing tomorrow whether or not annotations that point to concepts should be a thing.  Personally I think that non-defining relationships are the better solution there.

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
QuyenLy87: Hi @dmcgihtsdo, The annotation is actually a refset but has a different form. So it will extend the class "ReferenceSetMember". And because it does not extend SnomedComponent,so those properties cannot be hidden by using "@JsonView(value = View.Component.class)". That's why I have to hide them by JsonIgnoreProperties.

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
QuyenLy87: Hi @dmcgihtsdo, the annotation value should be @en"value", then you can retrieve the annotationLanguage

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
QuyenLy87: Hi @dmcgihtsdo, I'm not sure about the annotationTarget. But at the moment, we don't store it in ES and also it's not used by UI

QuyenLy87:
Hi @dmcgihtsdo, I've updated the implementation. Please have a look.
- Update the Annotation class. So some properties will be hidden (deleted, changed, and additionalFields, ...)
- Update unit test with conflict https://github.com/IHTSDO/snowstorm/blob/5a6774feabd42d0f7cbd2b424d3bcde547c145a7/src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java#L4587
- An example for the class:
`Annotation annotation = new Annotation();
		annotation.setReferencedComponentId("50960005");
		annotation.setModuleId("900000000000207008");
		annotation.setAnnotationTypeId("123456");
		annotation.setAnnotationValue("Testing");
		annotation.setAnnotationLanguage("en");`

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
```diff
@@ -4600,27 +4600,34 @@ void testRebasingAnnotationChanges() throws ServiceException, InterruptedExcepti
...
 
-		// Task will have one annotation after rabasing from project
-		pizza = conceptService.find(pizzaId, "MAIN/projectA/taskA");
-		assertTrue(pizza.getAnnotations().size() == 1);
+		// Rebase task B
+		MergeReview review = getMergeReviewInCurrentState("MAIN/projectA", "MAIN/projectA/taskB");
+		Collection<MergeReviewConceptVersions> conflicts = reviewService.getMergeReviewConflictingConcepts(review.getId(), new ArrayList<>());
+
+		// Expecting the annotation gets conflict
+		assertTrue(conflicts.size() != 0);
```
dmcgihtsdo, line 4630: @QuyenLy87 thanks for modifying this test. Can you expand it so you accept/finalise the conflict resolution? i.e. accept one of the columns and have a follow up assertion.

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,156 @@
...
+		}
+		return this;
+	}
+
+	public ReferenceSetMember toRefsetMember() {
+		String refsetId = getRefsetId() != null ? getRefsetId() : Concepts.ANNOTATION_REFERENCE_SET;
+		String annotationId = getAnnotationId() != null ? getAnnotationId() : UUID.randomUUID().toString();
+		ReferenceSetMember member = new ReferenceSetMember(annotationId, getEffectiveTimeI(), isActive(), getModuleId(), refsetId, getReferencedComponentId());
+		member.setAdditionalField(AnnotationFields.ANNOTATION_TYPE_ID, getAnnotationTypeId());
+		member.setAdditionalField(AnnotationFields.ANNOTATION_VALUE, getAnnotationLanguage() != null ? "@" + getAnnotationLanguage() + "\""  + getAnnotationValue() + "\"" : getAnnotationLanguage());
```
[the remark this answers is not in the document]
dmcgihtsdo: @QuyenLy87 Can you add some tests (possibly to ConceptSerialisationTest) so these faults are picked up automatically?

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
QuyenLy87: Hi @dmcgihtsdo, I've updated that test case. Please have a look. Thanks

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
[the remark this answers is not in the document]
QuyenLy87: The ConceptSerialisationTest may not detect this issue. But this test case will solve the issue https://github.com/IHTSDO/snowstorm/blob/a2738fa8fab4f9fbb970c22fc00ab58acf27e2a2/src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java#L511

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
dmcgihtsdo: Thanks @QuyenLy87. Do you know if the test will still pass if you change `conflict.getSourceConcept` to `concept.getAutoMergedConcept`? I am thinking an update will be needed in the `AutoMerger` class.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
```diff
@@ -265,6 +265,11 @@ PersistedComponents saveNewOrUpdatedConcepts(
 			// Axiom refset members
 			markDeletionsAndUpdates(newVersionConcept, existingConcept, existingConceptFromParent, Concept::getAllOwlAxiomMembers,
 					defaultModuleId, refsetMembersToPersist, rebaseConflictSave);
+			
+			// Annotation refset members
+			markDeletionsAndUpdates(newVersionConcept, existingConcept, existingConceptFromParent, Concept::getAllAnnotationMembers,
```
dmcgihtsdo, line 270: @QuyenLy87 when deleting a Concept with an annotation, the Concept is deleted but the annotation isn't. Is this expected?

src/test/java/org/snomed/snowstorm/core/data/services/BranchMergeServiceTest.java
QuyenLy87: Hi @dmcgihtsdo, Thanks for catching this issue. Much appreciated.
I've updated AutoMerger and the unit test. Please have a look.

src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java
QuyenLy87: Hi @dmcgihtsdo, Good catch. I've updated the logic. You can find the test case here
https://github.com/IHTSDO/snowstorm/blob/18db3d6c2bc27efcbc9ad71f38fb26a2c48f41f0/src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java#L236

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
```diff
@@ -0,0 +1,165 @@
...
+		} else {
+			setAnnotationValue(annotationValue);
+		}
+		return this;
+	}
+
+	public ReferenceSetMember toRefsetMember() {
+		String refsetId = getRefsetId() != null ? getRefsetId() : Concepts.ANNOTATION_REFERENCE_SET;
+		String annotationId = getAnnotationId() != null ? getAnnotationId() : UUID.randomUUID().toString();
+		ReferenceSetMember member = new ReferenceSetMember(annotationId, getEffectiveTimeI(), isActive(), getModuleId(), refsetId, getReferencedComponentId());
```
dmcgihtsdo, line 115: What's the expected payload when creating an Annotation @QuyenLy87? When creating a Concept without a module id, the default branch module is applied. When creating an Annotation without a module id, none is specified so null reaches the database.

src/main/java/org/snomed/snowstorm/core/data/domain/AnnotationView.java
```diff
@@ -0,0 +1,30 @@
+package org.snomed.snowstorm.core.data.domain;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.snomed.snowstorm.core.pojo.TermLangPojo;
+
+@JsonDeserialize(as = Annotation.class)
+public interface AnnotationView {
```
dmcgihtsdo, line 7: @QuyenLy87, the create Concept response is different from the read Concept response in that some properties of the Annotation are only returned when reading the Concept. If these properties are needed by the UI after creating the Annotation, a follow up read request will be needed.

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
QuyenLy87: Hi @dmcgihtsdo, Are you sure that you provided the default module for the branch? If yes, it will be used for Annotation while saving the concept.

src/main/java/org/snomed/snowstorm/core/data/domain/AnnotationView.java
QuyenLy87: Hi @dmcgihtsdo, what do you mean here?

src/main/java/org/snomed/snowstorm/core/data/domain/Annotation.java
dmcgihtsdo: Good spot @QuyenLy87, yes I hadn't supplied the default module. However, the default module isn't used for International, only Extensions.

QuyenLy87: Here is how we populate the default module ID if not specified for RefsetMember
https://github.com/IHTSDO/snowstorm/blob/develop/src/main/java/org/snomed/snowstorm/core/data/services/ConceptUpdateHelper.java#L290

QuyenLy87: Ah,, Yes, I see what you mean. I updated the Annotation class. If no module is provided, then use CORE_MODULE instead.

src/main/java/org/snomed/snowstorm/core/data/domain/AnnotationView.java
QuyenLy87: Thanks @dmcgihtsdo. I missed some pieces of code and just corrected them. Please have a look. Thanks

dmcgihtsdo:
All looks good now @QuyenLy87. Can you squash your commits / include the ticket reference when you go to develop?

dmcgihtsdo (approved):

Issue #365, opened 2022-01-16
ARM64 images missing in Docker Hub


metaruslan:
Hi guys!
https://hub.docker.com/r/snomedinternational/snowstorm/tags is missing ARM images. Having multi-arch builds is the recommended approach:
https://docs.docker.com/desktop/mac/apple-silicon/
There are 2 use cases:
1) Running on m1 macs (I see you do support m1 macs given this issue https://github.com/IHTSDO/snowstorm/issues/344 is already fixed in the code).
2) Running on AWS Gravitron (ARM) instances.
Is there any plan to provide ARM images? Is there a workaround for now for how to create an ARM image on an m1 mac manually?

Thanks in advance!

metaruslan:
I'm not familiar with the snowstorm maven docker setup, but here is what docker docs say:
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
https://github.com/docker/buildx#building-multi-platform-images
I haven't done it myself, so I cannot say what's the best approach. The end result is having linux/arm64/v8 in the OS/ARCH column of tags, e.g. like this:
https://hub.docker.com/_/elasticsearch?tab=tags

johngrimes:
I think the best way to do this is to break free of your Dockerfile and use Jib!

It’s easiest if you use a base image that is already multi-arch. It looks like you are using openjdk, so that should be fine.

An example of a multi-arch Jib build is in the Pathling POM: https://github.com/aehrc/pathling/blob/main/fhir-server/pom.xml

kaicode:
I like the sound of what @johngrimes mentioned but I have no expertise or capacity in this area at the moment.
@martingall87 yes please share your work if you get something working.

martingall87:
> I like the sound of what @johngrimes mentioned but I have no expertise or capacity in this area at the moment.
> @martingall87 yes please share your work if you get something working.

Hi @kaicode, I've had a look into using a multi-arch Jib build as @johngrimes suggested. It was fairly straight forward to setup with limited changes required. The changes I've made are available here:
- Branch - https://github.com/martingall87/snowstorm/tree/jib-arm64-docker-support
- Commit -https://github.com/martingall87/snowstorm/commit/6af8ba8b8ccfdb7c64735346fc5e44490dc03c39

I've tried to document the changes in `docs/using-jib-to-build-containers.md` and how to build images. I'm more than happy to have a discussion about it.

kaicode:
@martingall87 your work looks great.

[...]

PR #560, opened 2023-10-11
ISTO-105 Docker Hub and multi-arch fixes

kaicode:
The Snowstorm Docker image had been out of date on Docker Hub for a couple of releases. The build had stopped working. The plugin we were using for building docker images is no longer maintained.

As [suggested by John Grimes](https://github.com/IHTSDO/snowstorm/issues/365#issuecomment-1113959063) from CSIRO I have tested building Snowstorm docker images using a different maven plugin called Jib. I leveraged [work from the Snowstorm fork of Martin Gall](https://github.com/IHTSDO/snowstorm/issues/365#issuecomment-1383677203) (at EMIS UK) to achieve this.

The Jib plugin method successfully:
- Builds images for both amd64 and arm64 arch architectures
- Publishes to Docker Hub from the command line
- Martin Gall has already updated the Snowstorm documentation for this

**However**, when using Jib we will need to build and publish the Docker Hub images from Jenkins rather than relying on Docker Hub to run a build. Jenkins will need Docker Hub credentials for this.


CoderMChu (approved):

PR #568, opened 2023-11-09
MAINT-2384 Fix decimal serialisation in ECL index.

kaicode:
This fixes the way that decimal numbers are stored in the Snowstorm semantic index, that is used for ECL queries.

Previously concrete domain numbers like `0.00005` were stored in the index in scientific notation like `5.0E-4`. This causes ECL queries that use that number to throw an exception.

Unit tests are included that test:
- the serialisation and deserialisation of this number
- An ECL query using this number that previously failed when the format was bad

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
@@ -10,6 +10,7 @@ import org.springframework.data.elasticsearch.annotations.Field;
 import org.springframework.data.elasticsearch.annotations.FieldType;
 import org.springframework.data.elasticsearch.annotations.Setting;
 
+import java.text.DecimalFormat;
 import java.util.*;
 
 /**
@@ -299,10 +300,19 @@ public class QueryConcept extends DomainEntity<QueryConcept> implements FHIRGrap
 					builder.append(type);
 					builder.append("=");
 					for (Object value : attributes.get(type)) {
-						if (!(value instanceof String)) {
+						String valueString;
+						if (value instanceof String) {
+							valueString = (String) value;
+						} else {
 							builder.append("#");
+							if (value instanceof Double || value instanceof Float) {
+								// Maximum decimal places is 6 - See https://confluence.ihtsdotools.org/display/mag/Concrete+Domain+Decimal+Places+and+Rounding
+								valueString = new DecimalFormat("#0.0#####").format(value);
+							} else {
+								valueString = value.toString();
+							}
 						}
-						builder.append(value);
+						builder.append(valueString);
 						builder.append(",");
 					}
 					deleteLastCharacter(builder);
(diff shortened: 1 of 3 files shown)


src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
```diff
@@ -299,10 +300,22 @@ private static String serializeMap(Map<Integer, Map<String, List<Object>>> group
...
+								// Maximum decimal places is 6 - See https://confluence.ihtsdotools.org/display/mag/Concrete+Domain+Decimal+Places+and+Rounding
+								valueString = new DecimalFormat("#0.0#####").format(value);
+							} else {
+								valueString = value.toString();
+							}
 						}
-						builder.append(value);
+						if (valueString.contains(".")) {
+							System.out.println();
+						}
```
dmcgihtsdo, line 317: What's this block for @kaicode? It looks like a left over from some testing?

src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
```diff
@@ -299,10 +300,22 @@ private static String serializeMap(Map<Integer, Map<String, List<Object>>> group
...
+						if (value instanceof String) {
+							valueString = (String) value;
+						} else {
 							builder.append("#");
+							if (value instanceof Double || value instanceof Float) {
+								// Maximum decimal places is 6 - See https://confluence.ihtsdotools.org/display/mag/Concrete+Domain+Decimal+Places+and+Rounding
+								valueString = new DecimalFormat("#0.0#####").format(value);
+							} else {
+								valueString = value.toString();
+							}
```
dmcgihtsdo, line 313: This fix doesn't work for me @kaicode. I created a concrete concept on a local instance, but I'm still getting "5.0E-4 is not a valid number". Is this fix for future concept creations or should it work for historical data (my test concept) too?

src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
kaicode: This is left over from debugging. Will remove.

src/main/java/org/snomed/snowstorm/core/data/domain/QueryConcept.java
kaicode: The fix requires a rebuild of the semantic index. It will also work on new concepts without rebuilding.
`POST /admin/{branch}/actions/rebuild-semantic-index`
The semantic index rebuild should at the base of the branch tree (MAIN) and the change rebased into child branches.

For extensions the International Release branch should have a semantic index rebuild, then the extension working branches should be rebased. Snowstorm will block rebase of extension branches because the upgrade function should be used. However I noticed in the code that the upgrade function has been changed to block upgrading to the same international release again - effectively blocking data fixes traveling into extension branches. 

In the short term the block could be removed in a local Snowstorm version and the upgrade performed in each environment by pointing that local Snowstorm instance at each Elasticsearch environment.
In the medium term I would suggest removing this block from the Snowstorm API level and implementing it in UI logic instead.

kaicode: Thinking again - if this issue only exists in one extension then the semantic index rebuild only needs to happen there. It could be run on that working branch and rebased into child branches.

PR #578, opened 2024-01-24
ISTO-107 Fix loading LOINC 2.73 and above

kaicode:
This PR adds config that is picked up by the HAPI FHIR library so that the new LOINC zip entry file naming is used, enabling the latest LOINC packages to be loaded.
This change has already been tested today, see jira ticket for details.


CoderMChu (approved):
@kaicode It looks good and thanks for the fix.

PR #587, opened 2024-03-28
MAINT-2529 Startup RS type checks fail gracefully.

kaicode:
Refset type checks fail gracefully if the MAIN branch is locked during startup.


pgwilliams (approved):
Looks great, thanks Kai.

@tshird what do you think about adding a Datadog alert for seeing "Unable to verify reference set type configuration." in the logs?   Actually I'll put this in the ticket...

PR #593, opened 2024-05-21
MAINT-2452

dmcgihtsdo:
If an Axiom's OWL Expression is incorrectly back to front (i.e. parent -> child instead of child -> parent), an update to the Concept will result in the OWL Expression being partly fixed (i.e. child -> child), resulting in a Concept being its own ancestor.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/core/data/services/AxiomConversionService.java
+++ b/src/main/java/org/snomed/snowstorm/core/data/services/AxiomConversionService.java
@@ -61,11 +61,25 @@ public class AxiomConversionService {
 		AxiomRelationshipConversionService conversionService = setupConversionService(branchPath);
 		for (Concept concept : concepts) {
 			for (Axiom axiom : concept.getClassAxioms()) {
-				String owlExpression = conversionService.convertRelationshipsToAxiom(mapFromInternalRelationshipType(concept.getConceptId(), axiom.getDefinitionStatusId(), axiom.getRelationships(), true));
+				String owlExpression;
+				ReferenceSetMember referenceSetMember = axiom.getReferenceSetMember();
+				if (!axiom.isActive() && referenceSetMember != null) {
+					owlExpression = referenceSetMember.getAdditionalField(ReferenceSetMember.OwlExpressionFields.OWL_EXPRESSION);
+				} else {
+					owlExpression = conversionService.convertRelationshipsToAxiom(mapFromInternalRelationshipType(concept.getConceptId(), axiom.getDefinitionStatusId(), axiom.getRelationships(), true));
+				}
+
 				axiom.setReferenceSetMember(createMember(concept, axiom, owlExpression));
 			}
 			for (Axiom gciAxiom : concept.getGciAxioms()) {
-				String owlExpression = conversionService.convertRelationshipsToAxiom(mapFromInternalRelationshipType(concept.getConceptId(), gciAxiom.getDefinitionStatusId(), gciAxiom.getRelationships(), false));
+				String owlExpression;
+				ReferenceSetMember referenceSetMember = gciAxiom.getReferenceSetMember();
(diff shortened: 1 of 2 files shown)


src/main/java/org/snomed/snowstorm/core/data/services/AxiomConversionService.java
```diff
@@ -61,11 +61,25 @@ public void populateAxiomMembers(Collection<Concept> concepts, String branchPath
 		AxiomRelationshipConversionService conversionService = setupConversionService(branchPath);
 		for (Concept concept : concepts) {
 			for (Axiom axiom : concept.getClassAxioms()) {
-				String owlExpression = conversionService.convertRelationshipsToAxiom(mapFromInternalRelationshipType(concept.getConceptId(), axiom.getDefinitionStatusId(), axiom.getRelationships(), true));
+				String owlExpression;
+				ReferenceSetMember referenceSetMember = axiom.getReferenceSetMember();
+				if (!axiom.isActive() && referenceSetMember != null) {
+					owlExpression = referenceSetMember.getAdditionalField(ReferenceSetMember.OwlExpressionFields.OWL_EXPRESSION);
+				} else {
+					owlExpression = conversionService.convertRelationshipsToAxiom(mapFromInternalRelationshipType(concept.getConceptId(), axiom.getDefinitionStatusId(), axiom.getRelationships(), true));
```
dmcgihtsdo, line 69: @pgwilliams The problem comes from the **convertRelationshipsToAxiom** method. Whenever a Concept is being updated, it's Axioms are being re-generated. I've changed this so the Axioms are only re-generated if they are active, otherwise use what was there previously.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -2383,6 +2384,193 @@ void testCreateNewExtensionComponent_EffectiveTimeIsNull_DefaultModuleId() throw
 		assertEquals("11000172109", concept.getModuleId());
 	}
 
+	@Test
+	void testInvalidInactiveAxiomNotReactivated() throws ServiceException {
+		String intMain = "MAIN";
+		Map<String, String> intPreferred = Map.of(US_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED), GB_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED));
```
dmcgihtsdo, line 2390: @pgwilliams this tests inactive axioms are not re-activated accidentally.

src/test/java/org/snomed/snowstorm/core/data/services/ConceptServiceTest.java
```diff
@@ -2383,6 +2384,193 @@ void testCreateNewExtensionComponent_EffectiveTimeIsNull_DefaultModuleId() throw
...
+		// Assert inactive axiom hasn't changed
+		Axiom invalidAxiom = concept.getClassAxioms().stream().filter(cA -> cA.getAxiomId().equals(invalidAxiomId)).collect(Collectors.toList()).get(0);
+		assertEquals(20240301, invalidAxiom.getEffectiveTimeI());
+		assertEquals(backToFront, invalidAxiom.getReferenceSetMember().getAdditionalField("owlExpression"));
+	}
+
+	@Test
+	void testActivatingInvalidAxiomThrowsException() throws ServiceException {
+		String intMain = "MAIN";
+		Map<String, String> intPreferred = Map.of(US_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED), GB_EN_LANG_REFSET, descriptionAcceptabilityNames.get(PREFERRED));
```
dmcgihtsdo, line 2468: @pgwilliams this tests that if an invalid inactive axiom is reactivated, an exception is thrown. The exception bubbles up from a transitive closure problem.

PR #639, opened 2024-11-13
Fixes #638 ISTO-125 LOINC pagination fix.

### LINKED ISSUE #638: Pagination error for LOINC codes - IllegalStateException: Object builders can only be used once
I've run into a strange problem when trying to paginate over LOINC codes. I can successfully call http://localhost:8080/fhir/ValueSet/$expand?url=http://loinc.org?fhir_vs&count=1000&offset=1000, but as soon as the offset goes to 10,000 I get an error. So http://localhost:8080/fhir/ValueSet/$expand?url=http://loinc.org?fhir_vs&count=1000&offset=10000 would give "IllegalStateException: Object builders can only be used once".

The same end-point works fine for SNOMED codes, so http://localhost:8080/fhir/ValueSet/$expand?url=http://snomed.info/sct?fhir_vs&count=100&offset=300000 is working fine.

Any idea?

Thank you

kaicode:
This fixes a bug in the Snowstorm FHIR API where it's not possible to paginate beyond a 10,000 code offset. 

This regression was introduced during the Elasticsearch 8 upgrade last year. It is caused by the Spring Boot Elasticsearch 8 library not allowing queries to be reused.

Local testing shows that these changes fix the bug. I have also added result sorting based on code to ensure consistent pagination with different page sizes which makes testing easier.

This screenshot shows using a very large offset to a page with the last three LOINC codes:
Screenshot 2024-11-13 at 14 27 59

Performance is quite poor because with Elasticsearch we can not jump to an offset beyond 10k, we have to scroll through the results. The search-after token technique would solve this but it's not compatible with the FHIR API.

### DIFF
--- a/src/main/java/org/snomed/snowstorm/fhir/services/FHIRConceptService.java
+++ b/src/main/java/org/snomed/snowstorm/fhir/services/FHIRConceptService.java
@@ -4,7 +4,6 @@ import ca.uhn.fhir.jpa.entity.TermCodeSystemVersion;
 import ca.uhn.fhir.jpa.entity.TermConcept;
 import co.elastic.clients.elasticsearch._types.query_dsl.BoolQuery;
 import com.google.common.collect.Iterables;
-
 import org.hl7.fhir.r4.model.CodeSystem;
 import org.hl7.fhir.r4.model.CodeType;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


CoderMChu:
@kaicode Good catch. Thanks for fixing this.

Hiaslafson:
@dmcgihtsdo  Any news about the review? We would need the functionality Thanks :)

pgwilliams (commented):
Looks great, will include with PR #646 and test together.

PR #650, opened 2025-01-27
Add CZ to Snowstorm default configuration

kaicode:
This PR adds the correct default config for the CZ codesystem.

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -221,13 +221,14 @@ codesystem.config.SNOMEDCT-BE=Belgium Edition|11000172109|be|Belgian Federal Pub
 codesystem.config.SNOMEDCT-CA=Canadian English Edition|20621000087109|ca|Canada Health Infoway
 codesystem.config.SNOMEDCT-CAF=Canadian French Edition|20611000087101|ca|Canada Health Infoway
 codesystem.config.SNOMEDCT-CH=Swiss Edition|2011000195101|en|eHealth Suisse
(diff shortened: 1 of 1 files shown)


pgwilliams (approved):

src/main/resources/application.properties
```diff
@@ -221,13 +221,14 @@ codesystem.config.SNOMEDCT-BE=Belgium Edition|11000172109|be|Belgian Federal Pub
 codesystem.config.SNOMEDCT-CA=Canadian English Edition|20621000087109|ca|Canada Health Infoway
 codesystem.config.SNOMEDCT-CAF=Canadian French Edition|20611000087101|ca|Canada Health Infoway
 codesystem.config.SNOMEDCT-CH=Swiss Edition|2011000195101|en|eHealth Suisse
-codesystem.config.SNOMEDCT-DE=German Edition|11000274103|de|Bundesinstitut für Arzneimittel und Medizinprodukte, BfArM
+codesystem.config.SNOMEDCT-CZ=Czech Republic Edition|11000279109|cs|Ministry of Health, Czech Republic
```
dmcgihtsdo, line 224: @kaicode just to confirm `|cs|` isn't a typo instead of `|cz|`?

dmcgihtsdo: Seen your comment on INFRA-14489. `cs` is the language code and not `cz`.

kaicode: @dmcgihtsdo you are right, cs is the language code but this should be the country code. I've fixed, squashed the commits and rebased this branch.
No changes needed for dialect config yet, thanks for checking.

PR #694, opened 2025-09-01
ISTO-143 Fix ECL ancestor 10K limit.

kaicode:
This is an ECL fix for queries like "(>(^331101000221109.774160008))" where the subquery returns more than 10 thousand results.

See https://jira.ihtsdotools.org/browse/ISTO-143


CoderMChu:

PR #711, opened 2026-03-30
MAINT-2634 Add custom elision analyzer to support searching with apostrophes

dmcgihtsdo:


src/main/java/org/snomed/snowstorm/config/ElasticsearchConfig.java
```diff
@@ -162,9 +162,26 @@ protected void initialiseIndices(boolean deleteExisting) {
...
 		// Initialise Elasticsearch indices
+		String customElision = "custom_elision";
 		Map<String, Object> settings = new HashMap<>();
 		settings.put("index.number_of_shards", indexShards);
 		settings.put("index.number_of_replicas", indexReplicas);
+		settings.put("analysis", Map.of(
+				"filter", Map.of(
+						customElision, Map.of(
+								"type", "elision",
+								"articles", new String[] { "l","m","t","qu","n","s","j","d","c","jusqu","quoiqu","lorsqu","puisqu" },
```
kaicode: Are we using these French articles for all languages?
Consider using different articles for specific languages. It should be driven by configuration.

kaicode: I actually think this is the wrong place to put the analysers, because it's not possible to treat different languages differently. I think the correct place for this logic is in ConceptUpdateHelper doSaveBatchDescriptions, where the language specific folding happens.

src/main/java/org/snomed/snowstorm/core/data/services/DescriptionService.java
```diff
@@ -986,6 +986,8 @@ private String constructRegexQuery(String term) {
 				}
 				continue;
 			}
+
+			word = word.replaceAll("^[a-zA-Z]'", ""); // replace elisions (e.g. i'm, l'intervention etc)
```
kaicode: I think this should be language specific, driven by configuration. Both apostrophe types should be included as both will commonly be used in user searches.

Issue #651, opened 2025-02-05
UK Edition Loading Tips 🇬‍🇧


kaicode:
**Intro**
The UK Edition of SNOMED CT is requires some special handling. The package is larger than most and some of the packages have an unusual format where there are multiple concept files within one zip file.

**Tips**
The UK Edition works very well in Snowstorm. Follow these tips for smooth loading of the RF2 package.

1. Use the Monolith Snapshot package
    - This is clearly labeled in TRUD "SNOMED CT UK Monolith Edition, RF2: Snapshot". 
    - The download filename has the format `uk_sct2mo_39.4.0_20250115000001Z.zip`
    - Other distribution formats, for example with filename `SnomedCT_UKEditionRF_PRODUCTION20250115.zip`, will **not** load correctly.
2. Increase Snowstorm's max-terms count setting to 1 million
    - Use this setting: `--elasticsearch.index.max.terms.count=1000000`
    - This can be added at the end of the java startup command or included in a config file, see [Configuration Guide](https://github.com/IHTSDO/snowstorm/blob/master/docs/configuration-guide.md).
3. Assign plenty of memory when loading the package
    - We know the monolith loads into Snowstorm well in Docker environment with 12g of memory where Elasticsearch has 6g and Snowstorm has 4g.
    - These are the recommended Docker Compose changes:
```
services:
  elasticsearch:
  ...
    environment:
    ...
      - "ES_JAVA_OPTS=-Xms6g -Xmx6g"
    ...
    mem_reservation: 6g

  snowstorm:
  ...
    entrypoint: java -Xms4g -Xmx4g --add-opens java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED -cp @/app/jib-classpath-file org.snomed.snowstorm.SnowstormApplication --elasticsearch.urls=http://es:9200 --elasticsearch.index.max.terms.count=1000000
  ...
    mem_reservation: 4g
```
Make sure Docker's Overall Memory Limit is set to 12g or above.

Happy Snowstorming! 🇬‍🇧

tomaszandrzej:
As of March 2026, the UK monolith edition RF2 snapshot import fails when the maximum term count is set to 1000000.
It succeeds when the term count is set to 1500000. A lower value may also work, but I have not tested this.

AzazelSensei:
Opened https://github.com/IHTSDO/snowstorm/pull/724 with the UK monolith loading notes.

[...]

PR #724, opened 2026-09-02
docs: add UK Edition loading tips #651

AzazelSensei:
Adds the UK Edition loading notes from #651.

I pointed the old UK zip warning at the TRUD monolith snapshot, set `elasticsearch.index.max.terms.count` to 1500000 (1000000 fails on recent packages), and listed the Docker memory overrides against the current compose file.


kaicode (approved):
Looks good to me. Thank you!

PR #727, opened 2026-09-14
ISTO-182 Default config for Korean and Vet Edition

kaicode:
This configuration is missing from snowstorm and is causing unexpected behaviour in the public browser.
See INFRA-17966 for further details including module verification.


kaicode:
To be clear. I am not expecting this PR to go through to fix the INFRA. This change is just to fix the gap for the future and other users.

